PR #43, opened 2015-05-25
File API supposrt

johnpasquarello:
SoundFile objects can now be initialized with File API File objects. Which will be loaded and played as normal from there.


therewasaguy:
Merged, thank you @johnpasquarello !

I'd like to build off your exmaple to make it work with the [p5.dom](http://p5js.org/reference/#/libraries/p5.dom) library's [createFileInput](http://p5js.org/reference/#/p5/createFileInput), ``.dragOver()``, ``.dragLeave()`` and ``.drop()``

PR #46, opened 2015-05-27
Master

b2renger:
- add new .mp3 and .ogg files in examples/_files for testing. 
- add beat detection from fft example
- add offline peak detection example from http://tech.beatport.com/2014/web-audio/beat-detection-using-web-audio/


therewasaguy:
Thank you @b2renger this is exciting!!

I'm going to try to parse out the functionality that has to do with visualizing the output, from the functionality that has to do with analyzing the sound and returning some data, because I would like to add these features to the library.

b2renger:
You are right ! let me know when you have integrated it in the lib, and I'll be able to re-write the example with the new functionnality.

Plus since I know this code it will help me understand how to integrate new stuff :)

Should I un-pull request this or do we still merge and rebuild from there ?

therewasaguy:
Let's keep the PR open for now, more soon!

PR #53, opened 2015-05-30
mod to master.js and helpers.js

b2renger:
move masterVolume and getMasterVolume from helper.js to master.js
add overloading of setMasterVolume(target,time,timeFromNow)
add inline documentation to Master module


b2renger:
this one should be better :)

therewasaguy:
great!! Thanks @b2renger

PR #58, opened 2015-06-01
Adding Gain class

b2renger:
simple gain node added as a new module


therewasaguy:
nice, thank you @b2renger !

The only thing I'm not sure about is the connectSource method. This would be a new method name. We have some a method that serves a similar purpose called ``setInput`` for FFT and Amplitude. I'll change the name for consistency, hope that sounds good by you!

PR #60, opened 2015-06-23
Audioin

polyrhythmatic:
fixed the audioin enumeration issues

can now use the getSources function with a callback instead of using listSources. listSources is now deprecated.


therewasaguy:
nice, thanks @polyrhythmatic !

src/audioin.js
```diff
@@ -224,25 +225,43 @@ define(function (require) {
...
+   * Returns a list of available input sources. Chrome allows the 
+   * the user to access a list of possible media sources. Firefox 
+   * allows the user to choose from input sources in the permissions
+   * dialogue.
+   *
+   * @method  getSources
+   * @param  {Function} callback [a callback to handle the sources 
+   * when they have been enumerated]
+   * @return {[type]}            [description]
+   */
```
therewasaguy: @polyrhythmatic just a small note about documentation.

Param and return are formatted like this:

```{type} name description```

I found it a little confusing that docblockr puts  ``[ ]`` around type and description, because we don't actually want any square brackets there. The only place we use square brackets is around the name, and only to specify that it's an optional parameter. So if the callback here was optional, we could say 

```
  * @param  {Function} [callback] a callback to handle the sources 
  *                               when they have been enumerated
```

I'll update this for now. More info on p5 documentation [here](https://github.com/processing/p5.js/wiki/Inline-documentation#specify-parameters)

therewasaguy: Also, it'd be very useful to add an example here

polyrhythmatic:
Do we still need to merge the two branches?

Please test it out and let me know what you think!

src/audioin.js
polyrhythmatic: Ah ok, I can do this as well - let me know.

therewasaguy: awesome, thanks @polyrhythmatic! You could add an in-line example to the documentatino following [this guide](https://github.com/processing/p5.js/wiki/Inline-documentation#adding-example-code), and also put a sketch that you were using for testing into the /examples folder of this repo.

Have you had a chance to try this out in the Desktop Editor?

polyrhythmatic: @therewasaguy have not tested it out yet! I'll add that to the list

PR #62, opened 2015-06-25
Audio in

polyrhythmatic:


polyrhythmatic:
@therewasaguy let me know what you think of the examples etc. can't seem to get rid of the multiple request for microphone dialogs unless HTTPS is used. I'm okay with it since this is just an example and is chrome only. Let me know if you have any thoughts on how to eliminate them... its tied directly to the getusermedia function.

polyrhythmatic:
also, works with p5 editor although the editor does not prompt for mic access

polyrhythmatic:
@lmccart Lauren I added a new example and a new feature, will this automatically update in the documentation when it is rebuilt? Let me know if you need me to update

therewasaguy:
@polyrhythmatic this is great!

The [in-line documentation example](https://github.com/processing/p5.js-sound/pull/62/files#diff-6154ea4125d4002b5b8f64bb6b3270e8R253) will show up on the website when it's rebuilt.

There are a few steps involved to actually previewing the documentation as it would appear on the p5 website:
- clone the [p5.js repo](https://github.com/processing/p5.js)
  - npm install your dependencies, and add the new p5.sound to the /lib folder
  - run grunt yui to generate new documentation data. This appears as a data.json file in the /docs/reference folder...
- clone the [p5.js-website repo](https://github.com/processing/p5.js-website/)
  - replace the data.json file in the /reference folder with the one you just generated
  - p5.js-website needs a server that can run PHP, which the python simple server can't. I use the free [MAMP](https://www.mamp.info/en/) to run an Mac Apache MySQL PHP server on my mac.

I'd love to get grunt yui working for p5.sound so that we can preview p5.sound documentation independently, within this repo, without having to do all that stuff!

---

For the select_audioIn_source example, I'd rename test.js to sketch.js like the other examples, and add comments to the top of the file to explain what it's doing.

Do you think the callback to getSources should be its own function? I find that easier to read than an anonymous function so that's what I tend to do in my examples, but maybe that's subjective.

p5.sound example files can be added manually to the p5.js repo [here](https://github.com/processing/p5.js/tree/master/examples/addons/p5.sound).

I'm wondering if this example could be renamed "audioIn_Multiple_Sources," and be made a little bit less specific to Chrome. I think the ability to get multiple input sources at the same time is useful, and this example could demonstrate/explain a bit about how different browsers handle input selection.

Trying it out in other browsers...
- Desktop editor has a value for each input - cool! I'm trying to think of an easy way to show the value (when it exists) in your example, instead of "input array position: 3".
- Opera lets me select the input for each source.
- Firefox gets the `` "This browser does not support MediaStreamTrack.getSources()" `` error. Not sure of the best way to handle that, since FF won't tell how many sources there are. Maybe it could still do the callback, and if the result is undefined, the example could still create two AudioIns, which would let the user choose the source.

polyrhythmatic:
@therewasaguy Right - I'll focus more on performance in the editor. We can do something simple like

```
if(sourcelist && sourcelist[i].label){
    //add the names of the sources to the sketch instead of numbers
} else {
    //add the device numbers to the page
}
````
I'll also remove the anonymous function and replace it with a separate function. The only thing is it's still asynchronous so it may be confusing in a different way. Whats the overall philosophy with p5 and callbacks?

One other thing is - I think more people will be accessing the example from a browser, no?

I don't want to overcomplicate the example but I'll try reworking it.

PR #63, opened 2015-07-10
changed panner to use stereopanner node as default when the browser s…

polyrhythmatic:
…upports it


polyrhythmatic:
@therewasaguy would love to get your thoughts on this! I'm not sure where to go with the "inputchannels" method since it isn't really necessary for the stereopanner node. I know you use this method in some of the other functions

therewasaguy:
Thanks @polyrhythmatic, this is great!

I believe the inputchannels method had to do with whether to fade between the left and right channels of a stereo file, or to mono-sum it for panning. And if you have a mono file, should it be sent equally to left and right channels, or only to the left channel.

It might be that this method is obsolete, as long as it behaves the same way across browsers whether or not they've implemented the Web Audio StereoPanner.

polyrhythmatic:
@therewasaguy I think we need to keep the method for now - I'm not sure about how widely the stereopanner node is supported yet. I suppose we leave the method in place in the stereopanner version and have it do nothing.

would the tests reflect whether or not this is working correctly? Do we have a test for this? I've just been working with the example.

therewasaguy:
@polyrhythmatic there are no real tests for panning yet...that'd be a fun thing to work on!

PR #77, opened 2015-11-08
Fix typos in docs for Signal.prototype.scale

toolness:
I am an audio noob so I'm not sure if this is right, but it *seems* like a typo...


therewasaguy:
thanks @toolness ! Sorry I missed this, I've been in the process of revisiting signal math for p5.sound in general in response to our conversation. This is helpful in the meantime

PR #91, opened 2016-01-25
I fixed the envelope stuff, added an "AD ramp" feature, and added three examples

spiricom:
The new example show off the exponential envelopes. 
There is a new Ramp feature, which I think of as a simple AD envelope generator or slew generator. 
It has a rise time and a fall time. It will generate an exponential curve up or down to the desired value in approximately the rise or fall time set in the "setRampAD" function. 

You can trigger it simply with the .rampAD function - it takes a unit, time to start, and value to ramp to, and an optional value to end at. Then it draws an attack/decay envelope that decays as soon as the attack phase is finished. 

Or you can simply use the .ramp function, which will ramp to the given value at the given time, and will use the rise time if it's going up or the decay time if it's going down. 

I think I've gotten rid of most of the glitches. However, there is still a slight weirdness, in that if you set a .play in motion, and then send an early release command, then it doesn't cut off the envelope as you might expect. This is because the .play has turned off .isTriggered, so it ignores the new release time that has been sent afterwards, even though it would schedule the release for earlier. Fixing this seems like it would require more effort, like comparing whether the new time is earlier than a previously held time. Not sure it's necessary.

Thanks for the help in getting the git stuff straight :)
-Jeff


therewasaguy:
This is really exciting, thank you @spiricom ! Env is sounding better in general, the new examples sound fantastic and the new methods are fun to play with.

Merged on [this branch](https://github.com/processing/p5.js-sound/tree/newTone).

Let's keep working on it in the separate branch for just a bit longer to discuss/think thru the best way to incorporate all this new functionality. So far, the Env class just has a few methods documented [here](http://p5js.org/reference/#/p5.env). It's already more than [Processing Sound's Env](https://processing.org/reference/libraries/sound/Env.html)). Processing leans towards small API's because it's possible that people who visit the documentation are new to these concepts, and if there are too many methods, it can be unclear where to start.

So I'm wondering if there are places where we can consolidate / group functionality. Here are a few ideas:

* ``ramp`` / ``rampAD`` - would it make sense to combine these into one method called ``ramp`` that accepts either one or two values?

* ``setRampAD`` / ``setRampPercentages`` - What are the scenarios where people would want to tweak the ramp percentages?

* ``setADSR`` / ``set`` - After we talked, I'm thinking we *can* consolidate these like this: remove "sustainTime" from the constructor/``set``, and let it resurface here in the only method that needs it: ``play(unit, timeFromNow, sustainTime)``

* From our conversation, I got the sense you'd prefer something a bit more concise like ``setADSR(attackTime, decayTime, sustainPercentage, releaseTime)``. And then a separate method to ``setRange(max, min)`` where max defaults to 1, min defaults to 0. Does that sound like an improvement?

* In general, the ramp is starting to feel like its own parallel envelope system, because setting values for a "ramp" method won't impact the play/triggerAttack/triggerRelease methods. I'm trying to think how to bring them together a bit more. Could ``setADSR`` also fill the role of ``setRamp`` and ``setRampAD`` if it also sets the ``rampAttackTC`` and ``rampDecayTC`` based on the first two params? Then, the documentation for ``ramp`` can explain that it interprets either the attack (if one value is provided) or attack/decay (if two) as "time constants".

Just putting these ideas out for discussion—curious what you think! Feel free to respond here.

Whatever we decide, good documentation and examples will go a long way. All the documentation and inline examples that show up [on p5js.org](http://p5js.org/reference/#/p5.env) are auto-generated from the inline documentation. You can read more about that [here](https://github.com/processing/p5.js/wiki/Inline-documentation). I started to add documentation blocks for the new methods [like this](https://github.com/processing/p5.js-sound/commit/dd99cbe02bff719c06ecdaab2a44aba85f4d5e3a). Feel free to tweak. It'd be really useful to discuss some of the examples you've created into examples for the website. Examples that are formatted [inline like this](https://github.com/processing/p5.js-sound/blob/d7717291bce5935df2555f52c9aa52ead9eb191c/src/env.js#L35) will show up on the p5js.org website's documentation page for that class / method.

One issue I wasn't able to figure out is ``rampAD`` on Firefox—seems like a problem with Firefox, but I made a github issue for us here anyway:
https://github.com/processing/p5.js-sound/issues/93

Also made an issue for the .play / .triggerRelease bug that you mentioned: https://github.com/processing/p5.js-sound/issues/94

spiricom:
ramp / rampAD - would it make sense to combine these into one method called ramp that accepts either one or two values?

---  Yes! That's better. Can you edit that into there?

    setRampAD / setRampPercentages - What are the scenarios where people would want to tweak the ramp percentages?

--- none, let's not have them settable, and remove it from the API. Good idea.

    setADSR / set - After we talked, I'm thinking we can consolidate these like this: remove "sustainTime" from the constructor/set, and let it resurface here in the only method that needs it: play(unit, timeFromNow, sustainTime)

--- that's an interesting idea... I would find that more understandable, so I'm down with it if you don't think breaking some existing code is too bad.

    From our conversation, I got the sense you'd prefer something a bit more concise like setADSR(attackTime, decayTime, sustainPercentage, releaseTime). And then a separate method to setRange(max, min) where max defaults to 1, min defaults to 0. Does that sound like an improvement?

--- I like that.

    In general, the ramp is starting to feel like its own parallel envelope system, because setting values for a "ramp" method won't impact the play/triggerAttack/triggerRelease methods. I'm trying to think how to bring them together a bit more. Could setADSR also fill the role of setRamp and setRampAD if it also sets the rampAttackTC and rampDecayTC based on the first two params? Then, the documentation for ramp can explain that it interprets either the attack (if one value is provided) or attack/decay (if two) as "time constants".

-- Yes! that's good.

therewasaguy:
@spiricom I think we're all set now!

I kept the set method, and p5.Env constructor function, basically the same for backwards compatibility.

spiricom:
Awesome!

On Sun, Jan 31, 2016 at 9:33 PM, Jason Sigal <notifications@github.com>
wrote:

> @spiricom <https://github.com/spiricom> I think we're all set now!
>
> I kept the set method, and p5.Env constructor function, basically the same
> for backwards compatibility.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/processing/p5.js-sound/pull/91#issuecomment-177718467>
> .
>



-- 
Composer, musician, instrument designer
snyderphonics.com * @snyderphonics

PR #95, opened 2016-01-28
added a getCentroid() function to return the spectral centroid of an fft

spiricom:
I also tried to do the auto-documentation thing - I think I did it correctly.


therewasaguy:
awesome! Thank you, @spiricom !

PR #98, opened 2016-02-10
changed the "very low value" that the exponential functions replace z…

spiricom:
…ero with to a lower number, because it was creating audible noise in patches with many oscillators


therewasaguy:
thanks, good ear!

Issue #47, opened 2015-05-27
Discussion on polyphonic synth implementation


b2renger:
I tried to get started with some kind of modular polyphonic synth implementation.

I went for a : 
- a PolySynth class to deal with voices allocation, enveloppe and filter.
- an AudioVoice class which is a super class
- Two synths class inheriting from the AudioVoice class.

I seems all right, I'm really not sure it's the best design as I'm no js expert, but I'm struggling as I want to be able to manipulate params from custom synths, directly from the PolySynth class.

I will investigate further but now it's out for dicussion, check out the code below :

```javascript



var psynth;

function setup() {
  	
  createCanvas(800, 300);
  smooth();

  frameRate(25);

  psynth = new PolySynth(15,DetunedOsc );
  
  
}


function draw() {
  background(0);
}

function mousePressed(){

    var note = int(map(mouseX,0,width,60,84));
    var length = map(mouseY,0,300,0,5);
  
  	psynth.setAdsr(0.25,0.25,length,0.25);

    var d = int(random(1,12));
    DetunedOsc.setDetune(d);

  	psynth.setNote(note);
    psynth.play();
    
}

//////////////////////////////////////////////////////////////////////////////////////////////
// A typical synth class which inherits from AudioVoice class
function SquareVoice(){

  AudioVoice.call(this);

  this.osctype = 'square';
  this.oscillator = new p5.Oscillator(this.note,this.osctype);
  this.oscillator.disconnect();
  this.oscillator.start();

  this.oscillator.connect(this.filter);

  this.setNote = function(note){
    this.note = note;
    this.oscillator.freq(midiToFreq(note));
  } 
}
SquareVoice.prototype = Object.create(AudioVoice.prototype);  // browsers support ECMAScript 5 ! warning for compatibility with older browsers
SquareVoice.prototype.constructor = SquareVoice;

//////////////////////////////////////////////////////////////////////////////////////////////
// A second one
function DetunedOsc(){

  AudioVoice.call(this);

  this.osctype = 'sine';
  this.detune = 5;

  this.oscOne = new p5.Oscillator(midiToFreq(this.note),this.osctype); 
  this.oscTwo = new p5.Oscillator(midiToFreq(this.note)-this.detune,this.osctype);
  this.oscOne.disconnect();
  this.oscTwo.disconnect();
  this.oscOne.start();
  this.oscTwo.start();

  this.oscOne.connect(this.filter);
  this.oscTwo.connect(this.filter);

[...]

PR #103, opened 2016-03-07
Synths definitions and polyphony

b2renger:
This PR brings in reference to #47 :
- a way to design synth inheriting methods from the new  audiovoice.js class.
- a way to handle polyphony with the polysynth.js class.

an example is in the example folder and added in the inline documentation. We can start discuss if this implementation is solid and what are its limits, and we could probably improve comments.


therewasaguy:
@b2renger wow, this is exciting! I will spend some time playing with this and get back later this week with some feedback.

Some general thoughts to start:
- The AudioVoice class basically combines p5.Env and p5.Oscillator. I think the methods that get passed on to an Env or an Osc should share the same name. For example `attackPlay` should be `triggerAttack`.
- The AudioVoice has a Filter - how is this used?
- I like the idea of setParams. Which params should be exposed? We don't have many examples of an options object in p5, but this could be useful.
- Let's think of shorter, simpler examples for the inline documentation

More soon!

b2renger:
* Yes definitely I'll work in that way.
* the filter is use to filter dc offset, it's habit I have, maybe it not usefull. When combining several oscillator in an additive kind of synth it could be usefull, its not made to be manipulable, it's just there before the output. It could be named synthOut, this way you know you need to connect your dsp graph to it.
* for the setParams it's really up to the person who made the synth definition. If you make a synth with a biquad per voice it could be biquad coeffs ! or really anything else. I'll try to build a few synth definitions in a specific example or series of them. It could be worth a specific tuto somewhere.
* ok for shorter, I'll remove one synth definition, we don't need two synth for the inline example. But I think the code inside the classic p5js function is quite short and simple.

b2renger:
@therewasaguy 
I made a first revision based on your remarks. I also include the generated libs (not sure I should).
I was wondering about the setNote function, in the oscillator we have freq() to set the frequency I thought about mapping it, but I think it's a good choice to keep the setNote one 
- it feels more natural in a synth to talk about notes in the MIDI sense rather than frequencies.
- it's meant to be overriden in the synth definition it can be used to set frenquencies on several oscillators.

b2renger:
Since it worked right away I added another synth definition a classic additive synth.
I use setParams to pass an array of harmonics and the oscillator type that the kind of thing I had in mind.

b2renger:
@therewasaguy 
I cleaned up a bit the example, an made a dedicated one for additive synth, it features a few interesting stuff : 
- params are set from Dom elements
- use of specific functions to draw the amplitude of each voice.

It raises some use cases for getters like psynth.getCurrentVoice() , or psynth.getCurrentVoiceId(). 
It would probably be cleaner to have a gain node for the synthOut before the filter ?
And I should probably remove the oscillator from the audiovoice.js in my mind, I should remove the oscillator and replace it with a gain.

I noticed a bug that seems to be only on chrome, sometimes the env won't fire correctly, the sound just inflates then cuts. It doesn't seem to happen on firefox. I will look for a related bug or test case in the repo.

b2renger:
@therewasaguy @spiricom if one of you can help it would be great :)
I still get issues about env. I tried to switch to exponential envelopps but I still get some strange behavior.

Maybe I'm doing something wrong, in both examples I have two different behavior : 
- granular example : the envelop clicks on attacks and on releases. What I would need here would be a trapezoidal env, something like attack/length/decay.
- additivesynth : sometimes the sound inflates and then cuts with a click. I don't manage to find a pattern in this behavior so I don't really get where it comes from.

therewasaguy:
@b2renger Apologies for my delayed response. The new Additive and Granular examples are really exciting.

I still need to spend some quality time with this and think about how to present this new functionality. Extending AudioVoice feels on the more advanced side of the spectrum. I want to make sure that the basic default functionality will be clear to people who are new to sound and/or code.

One major question I have is whether AudioVoice and PolySynth are meant to be synths/oscillators, or are meant to be hacked to give voice to any sound sources (like the SoundFile in your Granular example). If it's the former, I think MonoSynth is a clearer name than AudioVoice, and this approach would simplify the API—we could initialize it with an Oscillator as you are currently doing. If it's the latter, we should keep the name consistent (i.e. MonoVoice / PolyVoice)*. I like the idea that the AudioVoice can extended in this way, but it definitely makes things more complicated. For example we might not want to initialize with an Oscillator, and what happens to all the Oscillator-specific methods like `setNote`? Or `freq()`, for that matter, which I think is only worth including if the default is to behave as a monosynth. I think setNote is definitely more useful, but don't want to impose an intonation)

*I'm not sure about the name "Voice" — people who are new to sound might think of the human voice, like this is a way to access the microphone. Open to keeping it but also open to other ideas, like I don't know if MonoSound / PolySound is any better?

b2renger:
@therewasaguy no pb for the delay :) 

First : 
> For example we might not want to initialize with an Oscillator, and what happens to all the Oscillator-specific methods like setNote? Or freq(), for that matter, which I think is only worth including if the default is to behave as a monosynth. I think setNote is definitely more useful, but don't want to impose an intonation)

I am for setNote as we talk about synth so we talk about music and we can't play a synth with frequencies. This method has to be overriden in the synth definition anyway to propagate the proper freq to the proper oscillator as it is done in the additive example. If we don't use oscillators it's just here and it can be confusing.

You told me "I have heard from other teachers that there is a need to have a very basic synth as part of the library" so lets keep that in mind.

This would drive me to gp for MonoSynth/PolySynth (+ the psynth as the name of variable in the example is a sweet bonus);  Maybe the extension of MonoSynth * should be documented, but building your own is probably difficult for beginners. 

To give some "sonic choices" to users we could ship a few synth definitions with the library but this will bring potentially a lot new classes to the api. Lets say 3 for now  : additive, fm and karplus strong :) ?

I mean we can keep the configuration that has been implemented but letting contributors deal with adding new synths to the lib. To keep it simple for methods we would need :  
* setNote 
* setADSR
* setParams(options[]) => for additive ( square, [1,2,5,7]) or  for fm (triangle, indexVal, ratioVal)
* play
* connect (to chain it to effects : delay, reverbs, filters this would give more sonic options)

the setParams method is tricky as the possible options change from synth definition to synth definition, so if we go for something like this it needs to be precisely documented.

Thus I would choose to exclude the use of soundFiles (on we use soundfiles in the classes we should have 'Sampler' mentionned) anyhow I think the sounfile class is pretty neat already and it should continue to hold all stuff specific to the use of samples. So I am in favor to add a new class that  would extend soundfile and deal with polyphony in its own way ... to be discussed later.

For the additive example as it is at this stage I was thinking about building a audio/visual synth for the gallery (having bloom from Brian Eno in my mind). A few options, a few clicks and a looping mechanism. What is interesting for me is the the use of amplitude object in each voice to track their level independantly and having graphic methods to display it by keeping on overriding audio voice :) that was pretty fun !

wow that's long !

therewasaguy:
Ok, I think this is coming together really well! Thanks for all you've done on this so far, it's gonna be really useful and lead to a lot of good-sounding p5 projects!

I just sent a pull request (and feel free to ignore if you've already done these things), but I wanted to make sure methods like play, triggerAttack and triggerRelease can be scheduled for precision. I also changed AudioVoice to MonoSynth.

I think it's good to set aside the Sampler idea for now to focus on the synths and to call them "Synths".

For PolySynth, should setNote, setParams and setADSR apply only to the current synth (using this.poly_counter), or to all synths? I would think all synths.

I'm not sure about setParams, there aren't options objects in other places in p5, so this really just seems useful for the more advanced examples. I'm thinking about the best way to get the word out about the examples...

spiricom:
Hi guys,
Cool work on this!
Sorry I haven't been piping in - just finishing up a vacation.
I'll try out these examples once I have a chance.
I'm going to be at the WebAudio conference at Georgia Tech in early April,
I'll let people know about this stuff :) Anyone else going?

-Jeff

On Fri, Mar 18, 2016 at 7:03 PM, Jason Sigal <notifications@github.com>
wrote:

> Ok, I think this is coming together really well! Thanks for all you've
> done on this so far, it's gonna be really useful and lead to a lot of
> good-sounding p5 projects!
>
> I just sent a pull request (and feel free to ignore if you've already done
> these things), but I wanted to make sure methods like play, triggerAttack
> and triggerRelease can be scheduled for precision. I also changed
> AudioVoice to MonoSynth.
>
> I think it's good to set aside the Sampler idea for now to focus on the
> synths and to call them "Synths".
>
> For PolySynth, should setNote, setParams and setADSR apply only to the
> current synth (using this.poly_counter), or to all synths? I would think
> all synths.
>
> I'm not sure about setParams, there aren't options objects in other places
> in p5, so this really just seems useful for the more advanced examples. I'm
> thinking about the best way to get the word out about the examples...
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/processing/p5.js-sound/pull/103#issuecomment-198572862>
>



-- 
Composer, musician, instrument designer
snyderphonics.com * @snyderphonics

therewasaguy:
Thinking about this more...Poly's `set*` methods seem like they should set all of the synths in the array. The `setNote` method makes sense for the MonoSynth, but for both poly/mono, how about `play`, `triggerAttack`, and maybe `triggerRelease` all accept a note to play/release as the first argument? This would more closely match the argument for p5.Env where the first is an object to control with the envelope.

So for example `monoSynth.play(midiVal, secondsFromNow, susTime);`

Worth noting this is similar to Tone.JS which has a [mono](http://tonejs.org/docs/#MonoSynth) and [polysynth](http://tonejs.org/docs/#PolySynth) class. In Tone, only one of each midi note value can be triggered at a time (like on a midi keyboard), so when you release, you have to specify which note you're releasing. I'm not sure if we want that or if that is the best solution to the difficult problem of specifying which note you want to release... at the moment, we don't have triggerRelease for PolySynth, but I think we need to be able to trigger attacks/releases to allow for interactivity.

@spiricom I'll be at WAC, look forward to seeing you there!

b2renger:
I corrected a bunch of stuff and now envelops work, it was my mistake :)

@therewasaguy I agree with you on the setNote, and passing the note to the play method makes sense.
for triggerAttack / triggerRelease could we think of something like a callback mechanism on the synth ? 

about the .setParams, I am not sure I understand how to do without it. We need a way to call it from polysynth that will instanciate a specific version of a monosynth, so we can't write specific methods in the a monosynth definition as they won't be accessible from psynth, or can we ?

the wac programm looks awesome, I'm jealous :) olos and constellation l

b2renger:
about the trigger Attack/Release of notes, I was thinking about interaction and I see several use cases here : 
1- a midi keyboard that we plugin with web midi and want to play it in a p5 project.
2- touch screen with touches[] and it's ids we could imagine a callback mechanism.
3- keyboard / makey the ways key events are handled doesn't allow key combination.

1 => will not be the current use case, until we get a webmidi lib ;) and even then in courses it requires some hardware, even if it you can find cheap decent midi controllers to experiment.
2 => I don't have a touchscreen on my laptop and even if more and more laptops come with touchscreen, it is still a hardware requirement. And then there is phones and tablets ... so it's probably worth exploring with MT in mind as a frame of reference.
3 => could we imagine having in p5js a keysPressed[] to identify which keys are down ? 

we could easily refine the polySynth class and add a boolean ('busy') in monoSynth to have some kind of 'synthStealing' mechanism keeping tonejs as an inspiration.

do you have any other interaction use case in mind ?

b2renger:
@therewasaguy I pushed some changes here if you have some time to check.

- dynamic voice allocation is on ! it uses the mechanism from Tone.js passing the the note value is an identifier in a map.
- removed setNote to pass note value through play instead
- I created new method name derived from the one we use in other objects : instead of tiggerAttack, triggerRelease I went for noteAttack and noteRelease as the syntax is not the same because we need to pass the note id 
- same goes for setADSR which is noteADSR. 

I have a bunch of small fixes to do, and I need to test noteAttack/noteRelease with key inputs but I think I'll be ready to create a few synth classes very soon.

Do I prefix the names of those new classes ? Synth (which would be a wrapper for an oscillator), SynthAdditive, SynthFM ?

b2renger:
Some thoughts after the latest changes

- about **setNote** : I'm not sure we shouldn't use setNote, in fact it's pretty handy. If you don't use setNote, the fact is that you have to override .play, .noteAttack, .noteRelease because if you have several oscillators you need to propagate the note parameter to every oscillator. If you have a setNote function you just overwrite this one, and you don't have to write new methods for the 3 use cases. It doesn't have to be called 'setNote' though, it could be something like :  '.midiNote'

- about **setParams** : I pretty sure it is the way to go. Maybe it's a bit unsual at first, but it's in fact really easy to use and to write (even for beginner) since every parameter has it's name written and its value. We probably just need good error messages for this.

- about **dynamic voices allocation** : it's pretty cool, but is there a way to know when an adsr is finished  because ? I'm not sure how to dispose of the voices created.

therewasaguy:
Hey @b2renger, nice work!

Here's a synths branch we can work off of https://github.com/processing/p5.js-sound/tree/synths

Had a chance to brainstorm about this with @spiricom a bit during WAC hack time.

- Re: dynamic voice allocation: A fixed number of allocated voices feels like the best solution. It could default to ~8 or so. Right now, `voices` is an object, we could keep it that way, or make it an array—either way we'll need to look up the synth by midiNoteVal when triggerRelease is called. But if we don't cap the number of voices it will become a memory issue. Each voice could be assigned a midiNoteVal when it is triggered, and we just rotate through them.

- For both setNote and setADSR, I think it's worth considering how everything in web audio is scheduled on the web audio clock. Those two parameters (or, at least, frequency and velocity) would be easier to schedule if they're incorporated into the play/triggerAttack method. Those methods could could accept a note, and a velocity ( using `p5.Env.setRange()` behind the scenes).

- Re `setNote`, I think that letting play/triggerAttack handle note values will make voice allocation easier. And I'm not sure I understand the advantage of `setNote`. Is the goal to set the same note to multiple oscillators? Or to have a more convenient way to set the note of more complex synths? In your example, it looks like you're already re-writing the triggerAttack and play methods to extend the base synth class. Maybe setNote is a semi-private method only for synth creators, so that we don't have to duplicate that code when creating a new synth.

- Let's also add a dispose method to both classes that disconnects and removes references to all of its resources.

Mind if I take a crack at some of this?

b2renger:
@therewasaguy a max number of voices is wise :)

The advantage of setNote is for complex synth, in the additive example you have a base note, and need to tune five oscillator based on multiplications of this base note. (in my example, I had to rewrite it to test it).

Indeed the paradigm of a pair of value note/velocity passed to the play method is nice. To replace setNote, we could have : '.createNote(pitch,vel)' 

Ok so now that it's a branch should we close the PR ?

therewasaguy:
Yes, feel free to work off of the the synths branch! I think you can push directly to it or do a PR if you want to review it first. I'll do the same. So far I've just updated it with the latest code from master but I'd be happy to give some of these ideas a shot.

I'm not sure about the idea of forming a note before playing it. Are there other examples of this in polyphonic synths?

One use case I'm thinking of is a song composition where I want to schedule three notes to play over the course of a few seconds. If I use createNote / setNote it would need to accept a time parameter:

` createNote(timeFromNow, pitch, vel)`
` play(timeFromNow)`

But this feels much more clear:
` play(timeFromNow, pitch, vel)`

I'm really leaning towards `_setNote` as a semi-private method that is called by triggerAttack and play, and synth developers can override it as they see fit. But the user doesn't need to see it, and synth developers wouldn't have to override both triggerAttack and play.

b2renger:
I'm not really aware of semi-private method and how they should be used. If you can create a small example that would be great, I'm all for it ! And I'll be able to continue with this new implementation.

My idea was to create the notes and just allocate them, and then play them whenever you want. The risk is that the note is being removed before it's played. So let's forget that.

PR #116, opened 2016-04-19
monoSynth tweaks

therewasaguy:
I made some tweaks to the MonoSynth:

- give it an `output` gain node for consistency with other p5.sound objects, and the p5.Env controls the output's gain.
- give it a default oscillator, in [this chain](https://github.com/processing/p5.js-sound/compare/synths...therewasaguy:synths?expand=1#diff-559e243463d4dac174c5eee32f469a81R42), which custom synth creators could override.
- create an [example for how to use it](https://github.com/therewasaguy/p5.js-sound/blob/860946847f8d6b207bf42331ef57a9813a8f88f7/examples/monosynth_basic/sketch.js)
- I found that using `env.ramp()` instead of `triggerAttack` and `triggerRelease` sounds much nicer (thank you @spiricom). Also it has the advantage of being able to schedule the velocity. If you schedule an attack in advance, you can schedule the velocity in advance, too—env.setRange does not have this ability and velocity is no longer part of triggerAttack so we can't schedule changes to the envelope velocity using `env.triggerAttack()`
- add connect, disconnect and dispose methods

@b2renger let me know if you have questions. If it looks good, wanna accept the pull request? I also started to take a look at PolySynth, and making sure that these revisions work with your custom PolySynth example, but will need to come back to it later.


b2renger:
It looks awesome ! let's do this !

PR #119, opened 2016-06-20
Fixed WAV format in soundRecorder.js:saveSound()

dorontal:
According to the WAV spec, the number that goes right after 'RIFF' is 36 + ... not 44 + ... - here's an excerpt from the spec:

    The canonical WAVE format starts with the RIFF header:
    0         4   ChunkID          Contains the letters "RIFF" in ASCII form (0x52494646 big-endian form).
    4         4   ChunkSize        36 + SubChunk2Size, or more precisely: 4 + (8 + SubChunk1Size) + (8 + SubChunk2Size) This is the size of the rest of the chunk following this number.  This is the size of the entire file in bytes minus 8 bytes for the two fields not included in this count: ChunkID and ChunkSize.


therewasaguy:
thank you @dorontal !

PR #124, opened 2016-07-25
Fix param misnaming (#1537)

nok:
Source: https://github.com/processing/p5.js/issues/1537
Error: https://github.com/processing/p5.js-sound/blob/master/src/env.js#L161-L162

Kudos to @michaelbach!

In addition I ran the default grunt task to rebuild the libraries.


therewasaguy:
thanks @nok and @michaelbach!

PR #132, opened 2016-08-26
Distortion effect based on WaveShaperNode

JoshuaStorm:
Currently generating the distortion curve based on a algorithm discussed in this [StackOverflow thread](http://stackoverflow.com/questions/22312841/waveshaper-node-in-webaudio-how-to-emulate-distortion)

Currently distortion `amount` values seem to make audible difference between 0-2000. 
Perhaps we should accept a more commonly accepted value range (0-127?) and scale from there to the 0-2000?

Currently the unit test is a bit light. Let me know if you have any ideas for a good test to add.


therewasaguy:
Nice! How about scaling 0-2000 to a value between 0 and 1?

JoshuaStorm:
Good call!

therewasaguy:
This is great, thank you @JoshuaStorm! I [tweaked / added some documentation](https://github.com/processing/p5.js-sound/commit/fa1744259b574594c1f3c8e8f9081cc540713b83), feel free to edit it further if you see areas where it could be improved.

I like the example, but people may not be familiar with the clean sound of a saw oscillator. I wonder if the example might work better with a soundfile and some kind of wet/dry control, like a slider that maps to distortion amount.

Issue #127, opened 2016-08-01
p5.Part.removePhrase(name) is not working 


lmccart:
reposting via @xievenlonger (https://github.com/processing/p5.js/issues/1538)

There's an error within p5.Sound that prevents the mentioned function from working. 

```javascript
    p5.Part.prototype.removePhrase = function (name) {
        for (var i in this.phrases) {
            if (this.phrases[i].name === name) {
                this.phrases.split(i, 1); // Should actually be SPLICE - than it works
            }
        }
    };
```

Sorry, I'm not sure how to create pull requests or anything :(

rjbultitude:
I'm happy to fix this @lmccart . I've already forked and have a branch ready for a pull request, so let me know whether someone else is on it or not.

therewasaguy:
@rjbultitude a PR would be great, thank you!

therewasaguy:
fixed with #133

PR #133, opened 2016-08-29
Bug fix for issue 127 removePhrase method

rjbultitude:
As per discussion in issue #127 @therewasaguy removePhrase has been updated to use splice rather than split on the phrases array.


rjbultitude:
@therewasaguy Should I add and commit lib/p5.sound.js and lib/p5.sound.min.js too having run grunt?

therewasaguy:
@rjbultitude we don't have a policy on including lib/ at the moment. It's definitely easier to see what was changed if we don't include the redundant build files, so this is good for now and it will be built into the next version of the library. Thanks for the fix/PR!

PR #142, opened 2016-10-16
Fix onended from triggering when SoundFile is paused

aksperiod:
For me, a Soundfile's `onended` method was being called on pause(), which according to the documentation should not be the case.

Fixed by ignoring the onended event if the soundfile is set to `_paused = true` because soundfile.stop() sets `paused` to `false`.

add `&w=1` to view it without the whitespace changes


therewasaguy:
thank you, @aksperiod ! Apologies for the delayer merge.

PR #147, opened 2016-10-29
Linlog averages

mkontogiannis:
This is my implementation of the linear and logarithmic groupings of the spectrum that we had discussed in the past. It also includes a function for generating the octave bands dynamically (https://courses.physics.illinois.edu/phys193/labs/octave_bands.pdf) and two examples respectively. 

I use them already for my browser based music visualisations, i.e: https://mkontogiannis.github.io/p5.js-sketches/frequency_bars/


therewasaguy:
Thank you @mkontogiannis this is super!

PR #159, opened 2017-01-08
Fixes mixingSound Example

TheFernDawg:
Replaced calls to connectSource with setInput.

This example was still calling p5.Gain.connectSource instead of
p5.Gain.setInput introduced in commit
41d6d1dccd8782410b460fcf7c938378efae898a causing the sketch not to run.


b2renger:
@TheFernDawg thank you !

PR #160, opened 2017-01-19
fixing https://github.com/processing/p5.js/issues/1739, 

lmccart:
using MediaDevices.enumerateDevices instead of getSources


therewasaguy:
thank you @lmccart !

PR #172, opened 2017-05-30
Effect.js draft

jvntf:
added effect.js and refactored filter.js to use inheritance


src/effect.js
```diff
@@ -0,0 +1,112 @@
+define(function (require) {
+	'use strict';
+
+	var p5sound = require('master');
+
+	p5.Effect = function(type) {
+		//everything points to the same audiocontext
+		this.type = type;
```
therewasaguy: Will all effects have `type` as the first argument in their constructor? `p5.Filter` already has this, but `p5.Reverb` and `p5.Delay` don't. I'd lean towards moving this to p5.Filter.

src/effect.js
```diff
@@ -0,0 +1,112 @@
...
+
+
+		// this.input.connect(this.output);
+		//connect input to wet and dry gain nodes
+		// this.input.connect(this.dry);
+		// this.input.connect(this.wet);
+
+		// //connect wet and dry gain nodes to output
+		// this.wet.connect(this.output);
+		// this.dry.connect(this.output);
```
therewasaguy: I know this is a work in progress PR, but just a reminder to clean up commented-out code before we merge. If you want to keep what you've started, you could work on wet/dry in a separate local branch.

src/effect.js
```diff
@@ -0,0 +1,112 @@
...
+
+		this.input.connect(this.output);
+
+		//connects to audiocontext destination
+		this.connect();
+
+		if (type) {
+			// determine the type of effect?
+			console.log(type);
+		}
```
therewasaguy: same comment as above—I think this belongs in the Filter constructor for now at least.

src/effect.js
```diff
@@ -0,0 +1,112 @@
...
+
+		//for now
+
+		this.dry.value = 1-value;
+		this.wet.value = 1;
+	};
+*/
+	//could link together a bunch of nodes
+	//TODO: find out how to have an open ended number of arguments
+	// p5.Effect.prototype.chain = function(...nodes){
```
therewasaguy: ooh, cool idea...  `...` is a spread operator that is new in ES6. Until we're able to transpile ES6 to ES5 for compatibility,  you could accomplish this with [`arguments`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments)

`chain` seems like it would be better off as a separate PR, once the baseline Effect class is done.

src/filter.js
```diff
@@ -73,33 +74,93 @@ define(function (require) {
...
-     *  <a href="http://www.w3.org/TR/webaudio/#BiquadFilterNode">
-     *  Web Audio BiquadFilter Node</a>.
-     *  
-     *  @property biquadFilter
-     *  @type {Object}  Web Audio Delay Node
-     */
+
+
+
+<<<<<<< HEAD
```
therewasaguy: It looks like there are some leftover markers from an unresolved merge conflict in this file. You'll have to pick through these markers manually. This is what they mean

```
<<<<<<< HEAD
This is the what you have locally
=======
```

```
=======
This is some conflicting commit
>>>>>>> a60f1128249aabc02c3d85ff12bf5f70d344f817```
```

Make sure to remove these markers keeping only the parts you want to keep, and when that's done it'll be easier for me to continue reviewing the changes to this file.

PR #173, opened 2017-06-01
Effect.js and modified Filter.js

jvntf:


src/filter.js
```diff
@@ -223,19 +213,6 @@ define(function (require) {
...
-   *  Disconnect all output.
-   *  
-   *  @method disconnect
-   */
-  p5.Filter.prototype.disconnect = function() {
-    this.output.disconnect();
-  };
 
   p5.Filter.prototype.dispose = function() {
     // remove reference from soundArray
```
therewasaguy: p5.Filter's dispose method can call `p5.Effect.prototype.dispose.apply(this)` and then just dispose of the pieces that are unique to this class.

For methods that extend superclass methods and have arguments, you can also pass in the `arguments` object like this:

`p5.Effect.prototype.method.apply(this, arguments)`

src/filter.js
```diff
@@ -73,33 +73,23 @@ define(function (require) {
...
-     *  The p5.Filter is built with a
-     *  <a href="http://www.w3.org/TR/webaudio/#BiquadFilterNode">
-     *  Web Audio BiquadFilter Node</a>.
-     *  
-     *  @property biquadFilter
-     *  @type {Object}  Web Audio Delay Node
-     */
+
+  p5.Filter = function (type) {
+    p5.Effect.call(this, 'Filter');
```
therewasaguy: I don't think we do anything with this 'Filter' argument, so it can probably be removed?

Also, a convention from other modules is to explicitly require dependencies. So we'd require p5 Effect at the top of this file, like `var Effect = require('effect')` and then refer to it as `Effect` within this module. That way this module isn't making any assumptions about whether p5.Effect already exists. With that approach we can remove `require('effect') from src/app.js as it'll get required by the modules that actually use it. I think the grunt task will result in similar output either way. Do you think that's a good approach?

src/filter.js
jvntf: cool! i wasn't sure if I could add to the Effect.prototype.dispose in objects that extended it.

src/filter.js
```diff
@@ -2,6 +2,7 @@ define(function (require) {
   'use strict';
 
   var p5sound = require('master');
+  var Effect = require('effect');
```
therewasaguy: You can change all "`p5.Effect`" in this module to "`Effect`" (once you add "return p5.Effect" to the Effect module)

src/effect.js
```diff
@@ -0,0 +1,56 @@
...
+		p5sound.soundArray.splice(index, 1);
+
+		this.input.disconnect();
+		this.input = undefined;
+
+		this.output.disconnect();
+		this.output = undefined;
+	};
+
+
```
therewasaguy: `return p5.Effect;` here so that Effect can be explicitly required by other modules

src/effect.js
```diff
@@ -0,0 +1,56 @@
...
+	p5.Effect.prototype.dispose = function() {
+		var index = p5sound.soundArray.indexOf(this);
+		p5sound.soundArray.splice(index, 1);
+
+		this.input.disconnect();
+		this.input = undefined;
+
+		this.output.disconnect();
+		this.output = undefined;
```
therewasaguy: don't forget to dispose of the wet/dry nodes here (or remove them from the constructor until we're ready to use them in a future PR)

It might also be good to clear the reference to the audio context (`this.ac = undefined`)

src/filter.js
```diff
@@ -73,33 +74,22 @@ define(function (require) {
...
-     *  <a href="http://www.w3.org/TR/webaudio/#BiquadFilterNode">
-     *  Web Audio BiquadFilter Node</a>.
-     *  
-     *  @property biquadFilter
-     *  @type {Object}  Web Audio Delay Node
-     */
+
+  p5.Filter = function (type) {
+	p5.Effect.call(this);
     this.biquad = this.ac.createBiquadFilter();
```
therewasaguy: check the spacing on line 81 (but really I should add linting to do this automatically)

examples/effect_test/Filter_LowPass/sketch.js
```diff
@@ -0,0 +1,78 @@
...
+  p = createP(description);
+  var p2 = createP('Draw the array returned by FFT.analyze( ). This represents the frequency spectrum, from lowest to highest frequencies.');
+}
+
+function draw() {
+  background(30);
+
+
+  // Map mouseX to a the cutoff frequency for our lowpass filter
+  filterFreq = map (mouseX, 0, width, 10, 22050);
```
therewasaguy: we could add constraint(0, 22050) to get rid of a warning message in the console when we try to set frequency above 22050

examples/effect_test/Filter_LowPass/sketch.js
```diff
@@ -0,0 +1,78 @@
...
+function setup() {
+  createCanvas(710, 256);
+  fill(255, 40, 255);
+
+  // loop the sound file
+  soundFile.loop();
+
+  //filter = new p5.LowPass();
+  
+  //filter = new p5.Filter();
```
therewasaguy: looks like unnecessary comments?

jvntf:
I have implemented the wet/dry feature via ToneJS CrossFade as well as a 'chain' feature in p5.effect. Added comments as well. Note: addition of wet/dry changes the way that effects are written-> effects that inherit from Effect.js must add a new processing node and connect nodes as follows: this.input -> this.some_effect -> this.wet.

lib/p5.sound.js
```diff
@@ -6771,11 +6769,20 @@ effect = function () {
     this.output.disconnect();
     this.output = undefined;
   };
+  p5.Effect.prototype.dispose = function () {
+    var index = p5sound.soundArray.indexOf(this);
+    p5sound.soundArray.splice(index, 1);
+    this.input.disconnect();
+    this.input = undefined;
+    this.output.disconnect();
+    this.output = undefined;
```
therewasaguy: Good thinking, @GoToLoop. I was under the impression that assigning to `undefined` was the best way to remove references for garbage collection. But [this article](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management) and [this one](http://bertanguven.com/preventing-memory-leaks-in-javascript-null-vs-delete) seem to suggest that assigning to `null` (rather than `undefined`) is the way to go?

GoToLoop: * Assigning _undefined_ or `null`, and even assigning anything else to the property, is fully enough for GC purposes.
* Operator `delete` is for removing the property itself from its object as well.
* If we execute `this.input = undefined;` or `this.input = null;`, and then `console.log('input' in this);`, it's still gonna print `true`!
* Only for when we `delete this.input;`, `console.log('input' in this);` is gonna print `false`.
* Therefore for 100% purge, use operator `delete` rather than simply assinging _undefined_ or `null` to the property.  ;-)

GoToLoop: * Just did a quick reading on the 2 articles you've linked to.  
* They're very old. And they still had fears about the very old IE non-compliant behaviors.
* Nowadays most of those glitches are gone, especially when activating `"use strict";`.
* The rule is simple: if there's only 1 property referencing an object, if we reassign that property w/ something else, or outright `delete` the property itself, the now unreferenced object is gonna be collected by the GC eventually.

therewasaguy: ah, thanks @GoToLoop ! Sounds like we should use `delete` for all dispose methods

PR #175, opened 2017-06-11
Change the fftSize on the fly!

TheBrenny:
Added a function that allows users to be able to change the bin size/fftSize.

The only issue is if the user inputs an invalid parameter, but I pretty much mimicked the way the constructor does it.

### DIFF
--- a/src/fft.js
+++ b/src/fft.js
@@ -91,27 +91,37 @@ define(function (require) {
    *  </code></div>
    */
   p5.FFT = function(smoothing, bins) {
-    Object.defineProperty(this, 'bins', {
-      get: function () {
-        return this.analyser.fftSize / 2;
+    this.input = this.analyser = p5sound.audiocontext.createAnalyser();
+    
+    Object.defineProperties(this, {
+      'bins': {
+        get: function() {
+          return this.analyser.fftSize / 2;
+        },
+        set: function(b) {
+          this.analyser.fftSize = (b * 2) || 1024;
+        },
+        configurable: true,
+        enumerable: true
       },
-      set: function (b) {
-        this.analyser.fftSize = b * 2;
-      },
-      configurable: true,
-      enumerable: true
+      'smoothing': {
+        get: function() {
+          return this.analyser.smoothing;
+        },
+        set: function(s) {
+          this.analyser.smoothingTimeConstant = s || 0.8;
+        },
+        configurable: true,
+        enumerable: true
+      }
     });
     
-    this.smoothing = smoothing || 0.8;
-    this.bins = bins || 1024;
-    var FFT_SIZE = bins*2 || 2048;
-    this.input = this.analyser = p5sound.audiocontext.createAnalyser();
+    //
+    this.smoothing = smoothing;
+    this.bins = bins;
 
     // default connections to p5sound fftMeter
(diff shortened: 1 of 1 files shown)


src/fft.js
```diff
@@ -472,6 +472,17 @@ define(function (require) {
...
+  
+  /**
+   * Changes the length of the resulting array, the bin size and in turn the fftSize of the analyser.
+   * 
+   * @method setBins
+   * @param {Number} bins    Must be a power of two between 16 and 1024.
+   */
+  p5.FFT.prototype.setBins = function(b) {
+    this.bins = b;
+    this.analyser.fftSize = b * 2;
```
therewasaguy: Similar to [this comment](https://github.com/processing/p5.js-sound/pull/176#discussion_r121283301), `this.bins` still has the potential to mismatch the `analyser.fftSize`, and that could be fixed with a getter.

and, I wonder if we need to add a new method for setBins, or if we could instead use a setter when we define the property?

TheBrenny: Can you show me an example of where `this.bins` might mismatch with `analyser.fftSize`? I mean, realistically the only way it can happen is if someone changes it by themselves, correct? 🤔

therewasaguy: Maybe this was addressed in #176, but the idea is that if I call

`fft.setBins(100)`

then this works

`this.bins = 100`

but this throws an error 

`this.analyser.fftSize = 100 * 2;`

`Failed to set the 'fftSize' property on 'AnalyserNode': The value provided (100) is not a power of two.`

TheBrenny: Ahhhh. I understand. I'll adjust the branch right now! Thank you for clearing that up! I've never really used this set of features of JS. 😅

src/fft.js
```diff
@@ -91,6 +91,17 @@ define(function (require) {
    *  </code></div>
    */
   p5.FFT = function(smoothing, bins) {
+    Object.defineProperty(this, 'bins', {
+      get: function () {
+        return this.analyser.fftSize / 2;
```
therewasaguy: I think we need to define `this.analyser` before we do `Object.defineProperty`

From there, we're good to go. And maybe we can do something like this

```
    var BINS = bins || 1024;
    var SMOOTHING = smoothing || 0.8;
    
    // create analyser
    this.input = this.analyser = p5sound.audiocontext.createAnalyser();

    // define properties
    Object.defineProperties(this, {
      'bins': {
        get: function() {
          return this.analyser.fftSize / 2;
        },
        set: function(b) {
          this.analyser.fftSize = b * 2;
        },
        configurable: true,
        enumerable: true
      },
      'smoothing': {
        get: function() {
          return this.analyser.smoothing;
        },
        set: function(s) {
          this.analyser.smoothingTimeConstant = s;
        },
        configurable: true,
        enumerable: true
      }
    });

    // now we can use the "setters"
    // to set the analyser and the p5.FFT props at the same time
    this.smoothing = SMOOTHING;
    this.bins = BINS;
```

Feel free to make that change or let me know if I should do it after merging these PR's. Thanks for inspiring these improvements, @TheBrenny !

TheBrenny: I'll close one PR and merge the two branches down to one and modify the remaining PR? :D

therewasaguy:
this is awesome, thank you @TheBrenny!

PR #176, opened 2017-06-11
Updated smooth function

TheBrenny:
Calling 

```javascript
var fft = new p5.FFT();
fft.smooth(0);
console.log(fft.smoothing);
console.log(fft.analyser.smoothingTimeConstant);
```

would yeild different results. This is because the smoothing time constant would get updated always, and `fft.smoothing` wouldn't because of the way that 0 is falsey (`0 == false`).

### DIFF
--- a/src/fft.js
+++ b/src/fft.js
@@ -91,6 +91,17 @@ define(function (require) {
    *  </code></div>
    */
   p5.FFT = function(smoothing, bins) {
+    Object.defineProperty(this, 'smoothing', {
+      get: function () {
+        return this.analyser.smoothingTimeConstant;
+      },
+      set: function (s) {
+        this.analyser.smoothingTimeConstant = s || 0.8;
+      },
+      configurable: true,
+      enumerable: true
+    });
+    
     this.smoothing = smoothing || 0.8;
     this.bins = bins || 1024;
     var FFT_SIZE = bins*2 || 2048;
@@ -99,7 +110,6 @@ define(function (require) {
     // default connections to p5sound fftMeter
     p5sound.fftMeter.connect(this.analyser);
 
-    this.analyser.smoothingTimeConstant = this.smoothing;
     this.analyser.fftSize = FFT_SIZE;
 
     this.freqDomain = new Uint8Array(this.analyser.frequencyBinCount);
@@ -468,7 +478,6 @@ define(function (require) {
    */
   p5.FFT.prototype.smooth = function(s) {
     this.smoothing = s;
-    this.analyser.smoothingTimeConstant = s;
   };
 
   p5.FFT.prototype.dispose = function() {
--- a/src/fft.js
+++ b/src/fft.js
@@ -467,9 +467,7 @@ define(function (require) {
    *                               Defaults to 0.8.
    */
   p5.FFT.prototype.smooth = function(s) {
-    if (s) {
-      this.smoothing = s;
-    }
+    this.smoothing = s;
     this.analyser.smoothingTimeConstant = s;
   };
 


src/fft.js
```diff
@@ -467,9 +467,7 @@ define(function (require) {
    *                               Defaults to 0.8.
    */
   p5.FFT.prototype.smooth = function(s) {
-    if (s) {
-      this.smoothing = s;
-    }
+    this.smoothing = s;
     this.analyser.smoothingTimeConstant = s;
```
therewasaguy: It is possible for `this.smoothing`  to be set even if `analyser.smoothingTimeConstant` fails.

One option is to switch the order.

But a better approach might be to to use a [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) for smoothing in the constructor of p5.FFT. What do you think of like this?

```
Object.defineProperty(this, 'smoothing', {
  get: function() {
    return this.fft.smoothingTimeConstant
  }
});
```

TheBrenny: I'm not sure if I fully understand what you're trying to explain... 🤔

Shouldn't `this.smoothing` be always equal to `this.analyser.smoothingTimeConstant`?

I've updated my comment to correctly align with the meaning that I had intended, but not correctly put across. :joy:

therewasaguy: @TheBrenny ah, sorry, it was confusing that I used `this.fft` in the example above when I should have used `this.analyser`. I'll try to clarify...

First of all, your comment is spot on that 0 is falsey and so your update is definitely an improvement. Anything beyond that is nice to have, let me know if you want to give it a shot.

I'm thinking, what if somebody calls `fft.smooth('some string');` ? Then in the smoothing function we'd see this go through successfully...

```this.smoothing = 'some string'```

...but this throw an error

`this.analyser.smoothingTimeConstant = 'some string'`

Same thing would happen if smoothing is above 1, or below 0.

So I am suggesting to take `this.smoothing` out of [the smooth() method](https://github.com/TheBrenny/p5.js-sound/blob/6de0770487c0dac22f085a4c81de7088122ec207/src/fft.js#L471), and instead define the property as a "getter" in the constructor. Something like this:

```
p5.FFT = function(smoothing, bins) {
    var SMOOTHING = smoothing || 0.8;

    ...
    this.analyser.smoothingTimeConstant = SMOOTHING;

    ...
    Object.defineProperty(this, 'smoothing', {
      get: function() {
        return this.analyser.smoothingTimeConstant;
      }
      set: function(s) {
         this.analyser.smoothingTimeConstant = s;
      }
    });
}
```

If we create an instance called fft, and set `fft.smoothing = 0`, that calls the setter, which does the same thing `fft.smooth(0)` would do. If we `console.log(fft.smoothing)` that will call the getter, returning `this.analyser.smoothingTimeConstant` in real time.

We haven't used getters and setters in p5.sound much, but seems like a useful place to incorporate them.

Comment (GoToLoop) [inline src/fft.js]: For parity reasons w/ properties created via operator `=`, when using **defineProperty()** to create an accessor property, include as well the descriptors _configurable_ & _enumerable_ set as `true`:

https://developer.Mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#Adding_properties_and_default_values

```js
p5.FFT = function (smoothing, bins) {
  this.analyser.smoothingTimeConstant = +smoothing || 0.8;

  Object.defineProperty(this, 'smoothing', {
    get: function ()  { return this.analyser.smoothingTimeConstant; },
    set: function (s) { this.analyser.smoothingTimeConstant = +s || 0.8; },
    configurable: true, enumerable: true
  });
};
```

TheBrenny: Literally the first time I've ever used getters and setters. Cheers for the instruction! 😄

src/fft.js
```diff
@@ -99,7 +110,6 @@ define(function (require) {
     // default connections to p5sound fftMeter
     p5sound.fftMeter.connect(this.analyser);
 
-    this.analyser.smoothingTimeConstant = this.smoothing;
```
therewasaguy: I think we still need to set this, and maybe we can do the `Object.defineProperty` below this line?

TheBrenny: Wouldn't it do the same thing? In the arrangement in the PR wouldn't it define the getters and setters, and then, in essence, set `this.analyser.smoothingTimeConstance = smoothing || 0.8`?

therewasaguy: oh yes, now this.smoothing does this, too, thanks to the setter!

src/fft.js
```diff
@@ -91,6 +91,17 @@ define(function (require) {
    *  </code></div>
    */
   p5.FFT = function(smoothing, bins) {
+    Object.defineProperty(this, 'smoothing', {
+      get: function () {
+        return this.analyser.smoothingTimeConstant;
```
therewasaguy: I think we need to create `this.analyser` before we define these properties

TheBrenny:
See #175

PR #178, opened 2017-06-13
Effects w inheritance

jvntf:
done: delay, reverb, distortion


therewasaguy:
this is looking good! Is it still in progress?

jvntf:
not in progress! But I have used continued to use `disconnect()` and `= 'undefined'` in `.dispose()` methods. I can update the dispose methods for the whole library to use `delete` in one PR

lib/p5.sound.js
```diff
@@ -1,4 +1,4 @@
-/*! p5.sound.js v0.3.3 2017-06-06 */
```
therewasaguy: For the next PR, let's hold off on adding the lib to commits—it'll make it easier to review because we can just focus on the `src/` and `test/` files.

src/delay.js
```diff
@@ -53,10 +55,7 @@ define(function (require) {
    *  </code></div>
    */
   p5.Delay = function() {
-    this.ac = p5sound.audiocontext;
-
-    this.input = this.ac.createGain();
-    this.output = this.ac.createGain();
+  	p5.Effect.call(this);
```
therewasaguy: `p5.Effect` works because of the order of app.js, but it's safer not to depend on that and to instead require dependencies explicitly. Try to use `Effect`, since that's what we required on line 6. Same goes for `Filter`, if you feel like cleaning up the lines that refer to `p5.Filter` (if you don't get to it in this PR, I will be happy to)

src/delay.js
```diff
@@ -261,52 +260,11 @@ define(function (require) {
...
-    var index = p5sound.soundArray.indexOf(this);
-    p5sound.soundArray.splice(index, 1);
 
-    this.input.disconnect();
-    this.output.disconnect();
+	 Effect.prototype.dispose.apply(this);
+
     this._split.disconnect();
     this._leftFilter.disconnect();
     this._rightFilter.disconnect();
```
therewasaguy: these are p5.Filters, so we should call dispose instead of disconnect to dispose of all its resources/references
```
this._leftFilter.dispose();
this._rightFilter.dispose();
```

src/distortion.js
```diff
@@ -118,38 +125,42 @@ define(function (require) {
...
-  p5.Distortion.prototype.disconnect = function() {
-    this.output.disconnect();
-  };
+  // /**
+  //  *  Send output to a p5.sound or web audio object
+  //  *
+  //  *  @method connect
+  //  *  @param  {Object} unit
+  //  */
+  // p5.Distortion.prototype.connect = function(unit) {
```
therewasaguy: hmm, we should probably keep documentation like this within each class/module, even though the methods are covered in p5.Effect module. These documentation blocks determine what shows up on https://p5js.org/reference/#/p5.Distortion methods section.

Another way to handle it might be to create documentation for p5.Effect, and then each class that inherits from p5.Effect includes a link in the documentation saying "inherits methods from p5.Effect".

src/reverb.js
```diff
@@ -115,44 +116,6 @@ define(function (require) {
...
-    var tFromNow = tFromNow || 0;
-    var now = p5sound.audiocontext.currentTime;
-    var currentVol = this.output.gain.value;
-    this.output.gain.cancelScheduledValues(now);
-    this.output.gain.linearRampToValueAtTime(currentVol, now + tFromNow + .001);
-    this.output.gain.linearRampToValueAtTime(vol, now + tFromNow + rampTime + .001);
-  };
-
-  /**
-   *  Send output to a p5.sound or web audio object
```
therewasaguy: more documentation that we want to keep accessible

src/delay.js
therewasaguy: actually I will merge and make this change in a separate PR

src/delay.js
therewasaguy: I'll make this change

src/delay.js
therewasaguy: ah, p5.Filter module needed to return/export the p5.Filter...

PR #179, opened 2017-06-13
Compressor

jvntf:
new effect node, compressor. not fully tested


src/compressor.js
```diff
@@ -0,0 +1,76 @@
...
+	};
+
+
+  p5.Compressor.prototype.set = function (attack, knee, 
+                                ratio, threshold, release) {
+
+    // var self = this;
+    // var errorTrace = new Error().stack;
+
+    if (attack) {this.attack(attack);}
```
therewasaguy: for all of these, `if (attack)` is a == comparison, and `0` will be "falsey". So if we want to allow for `0`, we should do a more explicit check like `if (attack) !== 'undefined'`

src/compressor.js
```diff
@@ -0,0 +1,76 @@
...
+    if (knee) {this.knee(knee);}
+    if (ratio) {this.ratio(ratio);}
+    if (threshold) {this.threshold(threshold);}
+    if (release) {this.release(release);}
+  };
+
+  p5.Compressor.prototype.attack = function (attack){
+    if (attack) {this.compressor.attack.value = attack;}
```
therewasaguy: I think all of these are [audioParam](https://developer.mozilla.org/en-US/docs/Web/API/AudioParam)s which means they can, in theory, be scheduled over time. Do we want to allow for this?

test/tests/p5.Compressor.js
```diff
@@ -0,0 +1,30 @@
...
+     it('wet dry value can be changed', function(){
+      var compressor = new p5.Compressor();
+
+      expect(compressor.drywet(0.5)).to.equal(0.5);
+    });
+
+    it('can set params', function(){
+      var compressor = new p5.Compressor();
+      compressor.set(0.5, 20, 15, -50, 0.75);
+      expect(compressor.attack(0.5)).to.equal(0.5);
```
therewasaguy: `compressor.attack()`

test/tests/p5.Compressor.js
```diff
@@ -0,0 +1,30 @@
...
+      compressor.dispose();
+    });
+
+     it('wet dry value can be changed', function(){
+      var compressor = new p5.Compressor();
+
+      expect(compressor.drywet(0.5)).to.equal(0.5);
+    });
+
+    it('can set params', function(){
```
therewasaguy: it might be good to test the `.set` method with incomplete params and ensure it sets the values appropriately

src/compressor.js
jvntf: i think we do! i'm adding a time option now-- 
but, this is one of the chunks of code that appears many times in the library, I wonder if it would be a worthwhile future project to make a base class for functions that are implemented specifically for many different object in p5-sound

therewasaguy: That's a great idea!

src/compressor.js
```diff
@@ -0,0 +1,146 @@
...
+   * Setter and Getter methdds 
+   * @method  {attack} set attack w/ time ramp or get current attack
+   * @method  {knee} set knee w/ time ramp or get current knee
+   * @method {ratio} set ratio w/ time ramp or get current ratio
+   * @method {threshold} set threshold w/ time ramp or get current threshold
+   * @method {release} set release w/ time ramp or get current release
+   */
+
+  p5.Compressor.prototype.attack = function (attack, time){
+    var self = this;
```
therewasaguy: The context isn't changing in these method, so we can just use `this` without creating the variable `self`. 

`var self = this` creates a reference to this instance of the compressor.

We would typically create a reference if we were referring to this compressor in a callback function or some other context.

src/compressor.js
```diff
@@ -0,0 +1,146 @@
...
+      self.compressor.threshold.linearRampToValueAtTime(threshold, self.ac.currentTime + 0.02 + t);
+    } else if (typeof threshold != 'undefined') {
+        threshold.connect(this.compressor.threshold);
+    }
+    return this.compressor.threshold.value;
+  };
+
+
+  p5.Compressor.prototype.release = function (release, time){
+    var self = this;
```
therewasaguy: just use `this`

src/compressor.js
```diff
@@ -0,0 +1,146 @@
...
+      self.compressor.ratio.linearRampToValueAtTime(ratio, self.ac.currentTime + 0.02 + t);
+    } else if (typeof ratio != 'undefined') {
+        ratio.connect(this.compressor.ratio);
+    }
+    return this.compressor.ratio.value;
+  };
+
+
+  p5.Compressor.prototype.threshold = function (threshold, time){
+    var self = this;
```
therewasaguy: just use `this`

src/compressor.js
```diff
@@ -0,0 +1,146 @@
...
+      self.compressor.attack.linearRampToValueAtTime(attack, self.ac.currentTime + 0.02 + t);
+    } else if (typeof attack != 'undefined') {
+        attack.connect(this.compressor.attack);
+    }
+    return this.compressor.attack.value;
+  };
+
+
+  p5.Compressor.prototype.knee = function (knee, time){
+    var self = this;
```
therewasaguy: just use `this`

src/compressor.js
```diff
@@ -0,0 +1,146 @@
...
+   * @param {Number} release   [default = .25, range 0 - 1]
+   */
+  p5.Compressor.prototype.set = function (attack, knee, 
+                                ratio, threshold, release) {
+
+    if (attack != 'undefined') {this.attack(attack);}
+    if (knee != 'undefined') {this.knee(knee);}
+    if (ratio != 'undefined') {this.ratio(ratio);}
+    if (threshold != 'undefined') {this.threshold(threshold);}
+    if (release != 'undefined') {this.release(release);}
```
therewasaguy: These type checks should be `typeof attack !== 'undefined'` with `typeof` and double equals.

Might be good to test that these values can be undefined.

This seems like a good candidate for an options object, but I think the API is good for now.

src/compressor.js
```diff
@@ -0,0 +1,146 @@
...
+      self.compressor.knee.linearRampToValueAtTime(knee, self.ac.currentTime + 0.02 + t);
+    } else if (typeof knee != 'undefined') {
+        knee.connect(this.compressor.knee);
+    }
+    return this.compressor.knee.value;
+  };
+
+
+  p5.Compressor.prototype.ratio = function (ratio, time){
+    var self = this;
```
therewasaguy: just use `this`

src/compressor.js
```diff
@@ -0,0 +1,146 @@
...
+	'use strict';
+
+	var p5sound = require('master');
+	var Effect = require('effect');
+  var CustomError = require('errorHandler');
+
+
+  /**
+   * Built on Web Audio Dynamics Compressor Node
+   * https://www.w3.org/TR/webaudio/#the-dynamicscompressornode-interface
```
therewasaguy: This merits a more details explanation and an example. That could be done in a future PR but let's make sure to provide some information about what a compressor is, and what all these parameters means. It is a difficult concept to explain, but an interactive example should help :)

test/tests/p5.Compressor.js
```diff
@@ -0,0 +1,29 @@
...
+      var compressor = new p5.Compressor();
+      compressor.set(0.5, 20, 15, -50, 0.75);
+      expect(compressor.attack()).to.equal(0.5);
+      expect(compressor.knee()).to.equal(20);
+      expect(compressor.ratio()).to.equal(15);
+      expect(compressor.threshold()).to.equal(-50);
+      expect(compressor.release()).to.equal(0.75);
```
therewasaguy: it would be good to test

- what happens when parameters are `undefined`
- what happens when parameters are invalid (we should expect an error)

src/compressor.js
```diff
@@ -0,0 +1,146 @@
...
+
+
+  p5.Compressor.prototype.ratio = function (ratio, time){
+    var self = this;
+    var t = time || 0;
+    if (typeof ratio == 'number'){
+      self.compressor.ratio.value = ratio;
+      self.compressor.ratio.cancelScheduledValues(self.ac.currentTime + 0.01 + t);
+      self.compressor.ratio.linearRampToValueAtTime(ratio, self.ac.currentTime + 0.02 + t);
+    } else if (typeof ratio != 'undefined') {
```
therewasaguy: !==

src/compressor.js
```diff
@@ -0,0 +1,146 @@
...
+
+
+  /**
+   * Built on Web Audio Dynamics Compressor Node
+   * https://www.w3.org/TR/webaudio/#the-dynamicscompressornode-interface
+   *
+   * @class p5.Compressor
+   * @return {Object} Compressor Object
+   */
+	p5.Compressor = function() {
```
therewasaguy: Should the constructor accept the same arguments as `set` (and call `set`)?

src/compressor.js
therewasaguy: (by the way, linting will help catch this kind of thing and suggest best practices, so maybe we should merge that PR!)

therewasaguy:
Can you `git rm examples/files/beat.mp3.asd` in your next commit?

src/compressor.js
jvntf: I don't have a problem with this! I didn't include `set` parameters just to match what seemed to be standard in other effects

test/tests/p5.Compressor.js
jvntf: should we be throwing an `Error()`? i think `compressor.method()` tests an `undefined` parameter (and it just returns the current value). currently passing in an invalid parameter yields the same result as `undefined`

PR #183, opened 2017-06-17
EQ.js

jvntf:


src/eq.js
```diff
@@ -0,0 +1,261 @@
...
+  /**
+   *  @{method} setBand Make adjustments to individual bands of the EQ
+   *
+   *  @{param} band {number} Band to be modified
+   *  @{param} option {string} Modify function (toggle, mod, type);
+   *  @{param} param1 {number} Set the frequency of a band w/ usage: "mod"
+   *  @{param} param2 {number} Set the Q value (resonance) of a band w/usage: "mod"
+   *  @{param} param1 {string} Set the type of the band filter w/ usage: "type"
+   */
+  p5.EQ.prototype.setBand = function (band, option, param1, param2) {
```
jvntf: I do agree that EQ should do as little re-implementation as possible-- The problem that I have run into with using [p5.Filter](https://p5js.org/reference/#/p5.Filter) to control the bands is that the are the [BiquadFilters](https://www.w3.org/TR/webaudio/#the-biquadfilternode-interface) are wrapped in [Gain](https://www.w3.org/TR/webaudio/#GainNode) nodes (due to [p5.Effect](https://github.com/processing/p5.js-sound/blob/master/src/effect.js#L10) and the input / output standard format of p5.js-sound.

All these issues extra Gain nodes seem to affect the accuracy of the EQ (boosting the overall volume, making it impossible to filter out highs...etc). I think this problem could be addressed either with different methods or by using the design you have indicated above, but with methods such as `.freq()` implemented in `p5.EQ`(because they have to work on the raw biquad filter). what do you think?

Gruntfile.js
```diff
@@ -97,7 +97,9 @@ module.exports = function(grunt) {
             'effect': 'src/effect',
             'filter': 'src/filter',
             'reverb': 'src/reverb',
+            'eq': 'src/eq',
             'distortion': 'src/distortion',
+            'compressor': 'src/compressor',
```
jvntf: sorry about this! must have merged a branch by accident-- will commit this compresser stuff out of here

src/eq.js
```diff
@@ -0,0 +1,261 @@
...
+   *  }
+  * }
+   *  </code></div>
+   */
+
+
+  p5.EQ = function(_eqsize) {
+
+    //default size is 8 band EQ
+    _eqsize = _eqsize || 8;
```
jvntf: this is a great point! I think the second part also addresses your next comment re: the approach to calculating frequencies and and Q values-- The values I used were relatively arbitrary just to get visual results for testing. Will do some research about standards / default values.

src/eq.js
therewasaguy: Hmm...

I wonder if these inaccuracies are something that could (should?) be fixed within p5.Filter.

Another approach might be to modify/extend p5.Filter after importing it into the p5.EQ module. I assume you already you try `disconnect()` to disconnect it from master? Maybe there is some additional patching that could be done, like setting filter.input and filter.output to the biquadFilternNode and deleting the gain nodes.

And/or, could it be worth abstracting out a p5.SimpleFilter class that gives an API that both p5.EQ and p5.Filter could inherit?

Just brainstorming...

jvntf: we could remove the gain nodes from p5.Filters-- but ultimately the difference in the API is that users would modify bands of the EQ either by 
- `eq.bands[i].set() or .freq() or .res()` <- the established filter API
or
- `eq.setBand(i,"mod",-40)` <- to change the gain of the ith band to -40
to me, it seems more straightforward to use a new .setBand() method from the highest level, rather than requiring users to access elements of the array of band filters. It might still be good to use p5.Filters internally, just to make things clearer for anyone looking at the library.

therewasaguy: If filters are accessed directly via the array, it would make sense to be consistent with array storage. I'm still leaning towards something like this, but we can chat more tomorrow.

jvntf:
because this PR has become very long, here is a quick summary of what changes it includes

- new `p5.EQ` class can construct an EQ of size 3 or 8
- abstracted `p5.Filter` of type peaking, with gain nodes deleted is used for each band
-added 2 methods to p5.Filter:
     - `.gain()` modifies the `gain` audio param of a BiquadFilterNode directly [the existing `.amp()` modifies the output gain node which yields a different result]
     - `.toggle()` switches the filter's type from whatever it is to 'allpass' and back again

src/eq.js
```diff
@@ -0,0 +1,103 @@
...
+   * @example
+   * <div><code>
+   * 
+   * </code></div>
+   */
+  p5.EQ = function(_eqsize) {
+    Effect.call(this);
+
+    //p5.EQ can be of size (3) or (8), defaults to 3
+    _eqsize = _eqsize == 3 || _eqsize == 8 ? _eqsize : 3;
```
therewasaguy: the convention is to always use "strict equality" `===` `!==` rather than "abstract equality" `==` `!=`

src/eq.js
```diff
@@ -0,0 +1,103 @@
...
+    src.connect(this.input);
+  };
+
+  /**
+   * Add a new band. Creates a p5.Filter and strips away the gain node wrapper
+   * @param  {[type]} freq [description]
+   * @param  {[type]} res  [description]
+   * @return {[type]}      [description]
+   */
+  p5.EQ.prototype.newBand = function(freq, res) {
```
therewasaguy: nice! We might want to rename this to _newBand to indicate that it is a private method. This can also be included in the yui documentation block http://yui.github.io/yuidoc/syntax/#private

src/filter.js
```diff
@@ -99,6 +99,9 @@ define(function (require) {
     if (type) {
       this.setType(type);
     }
+
+    this._on = true;
+    this.type = this.biquad.type;
```
therewasaguy: At first glance, it wasn't clear why there is a duplicate property. Now I see how this is useful for `toggle()`. Maybe rename to `this._untoggledType` to indicate that it's a private property and describe how it is used?

What happens if a user calls `setType()` on the filter itself, and then `toggle`? It'd be good to write a test case for this and make sure it can pass. [Getters and setters](https://javascriptplayground.com/blog/2013/12/es5-getters-setters/) might be useful here.

test/tests/p5.EQ.js
```diff
@@ -0,0 +1,62 @@
...
+      expect(eq.bands.length).to.equal(8);
+      eq.dispose();
+      eq = new p5.EQ(50);
+      expect(eq.bands.length).to.equal(3);
+      eq.dispose();
+    });
+
+    it('a band can be toggled on and off', function() {
+      var eq = new p5.EQ(8);
+      eq.setBand(2, "toggle");
```
therewasaguy: update the tests for the new API

examples/Filter_BandPass/sketch.js
```diff
@@ -15,7 +15,8 @@ function setup() {
   createCanvas(710, 256);
   fill(255, 40, 255);
 
-  filter = new p5.BandPass();
+  filter = new p5.Filter('peaking');
+  filter.biquad.gain.value = -40;
```
therewasaguy: did you mean to create a separate example for this?

src/eq.js
```diff
@@ -0,0 +1,104 @@
...
+    var freq, res;
+    for (var i = 0; i < _eqsize; i++) {
+      if (i == _eqsize - 1) {
+        freq = 20480; 
+        res = .1;
+      } else if (i == 0) {
+        freq = 160;
+        res = .1;
+      } else {
+        freq = this.bands[i-1].freq * factor;
```
therewasaguy: this line was returning NaN, but calling the `freq()` function does the trick—nice!

`        freq = this.bands[i-1].freq() * factor;`

examples/eq/sketch.js
```diff
@@ -0,0 +1,132 @@
...
+  noise.disconnect();
+  eq.process(noise);
+  noise.start();
+
+  //use an fft to visualize the sound
+  fft = new p5.FFT();
+
+  for (var i = 0; i < eqSize; i++) {
+    cntrlPts[i] = new CntrlPt(i, (width/(eqSize-1)) * i, height/2);
+    eq.bands[i].freq(map(cntrlPts[i].x, 0, width, 160, 20480));
```
therewasaguy: I think this would be a stronger example if you used the default EQ positions—highlight the exponential scaling that you worked on.

Some nice to haves
- Labels for the nodes showing their frequency and gain
- It'd be really cool to scale the frequency spectrum logarithmically. There's so much good stuff at the lower end of the frequency spectrum, every octave gets equal number of pixels on the screen that gives the user more room to play with meaningful frequencies. [Here's](https://therewasaguy.github.io/p5-music-viz/demos/04_fft_freq_spectrum/) an attempt I made at log/lin toggle for FFT visualization, in case it's helpful

jvntf:
@therewasaguy finished the example for EQ, and made some changes to the way the default spectrum is laid out

test/tests/p5.EQ.js
```diff
@@ -0,0 +1,72 @@
...
+      expect(eq.bands[2].gain()).to.equal(30);
+    });
+
+    it('a bands freq value can be changed', function() {
+      var eq = new p5.EQ(8);
+      console.log(eq);
+
+
+
+      expect(eq.bands[0].freq()).to.equal(160);
```
therewasaguy: this is throwing an error, should it be 100 or 160?

test/tests/p5.EQ.js
```diff
@@ -0,0 +1,72 @@
+define(['chai'],
+  function(chai) {
+
+  var expect = chai.expect;
+
+  describe('p5.EQ', function() {
+
+    it('can be created and disposed', function() {
+      var eq = new p5.EQ();
+      console.log(eq);
```
therewasaguy: good practice to remove these console.log's

src/eq.js
```diff
@@ -0,0 +1,138 @@
...
+
+
+    // this.bands[0] = this._newBand(0,.1);
+
+    // this.input.connect(this.bands[0].biquad);
+    // this.bands[_eqsize-1] = this._newBand(22050, .1);
+    // for (var i = 1; i < _eqsize-1; i++) {
+    //   this.bands[i] = this._newBand(344 * Math.pow(2, i-1), .1);
+    //   this.bands[i-1].biquad.connect(this.bands[i].biquad);
+    // }
```
therewasaguy: can we delete this comment?

src/eq.js
```diff
@@ -0,0 +1,138 @@
...
+   *  
+   * @class p5.EQ
+   * @constructor
+   * @param {Number} [_eqsize] [Constructor will accept 3 or 8, defaults to 3]
+   * @return {Object} [p5.EQ object]
+   *
+   * @example
+   * <div><code>
+   * 
+   * </code></div>
```
therewasaguy: this results in an empty example in the documentation

screenshot 2017-07-24 23 47 13

src/eq.js
```diff
@@ -0,0 +1,138 @@
...
+   * This method adds a new band to the EQ's array of bands.
+   * By default the method will set the new band to have a 
+   * frequency of 20480 and a resonance of .1. These values can be modified
+   * after the band is created.
+   *
+   * @method addBand
+   * 
+   */
+  
+  p5.EQ.prototype.addBand = function(freq, res) {
```
therewasaguy: Do we need this method? Or do we need a removeBand method, too?

src/eq.js
```diff
@@ -0,0 +1,138 @@
...
+    Effect.call(this);
+
+    //p5.EQ can be of size (3) or (8), defaults to 3
+    _eqsize = _eqsize === 3 || _eqsize === 8 ? _eqsize : 3;
+
+    var factor;
+    _eqsize == 3 ? factor = Math.pow(2,3) : factor = 2;
+
+    //bands are stored in an array, index 0 - 3 or 0 - 7
+    this.bands = [];
```
therewasaguy: This needs documentation. "An array of modified p5.Filter objects," highlight their freq and gain methods, and provide a couple examples?

src/eq.js
```diff
@@ -0,0 +1,157 @@
...
+   * @param {Number} [freq3] Frequency value for band with index 3
+   * @param {Number} [gain3] Gain value for band with index 3
+   * @param {Number} [freq4] Frequency value for band with index 4
+   * @param {Number} [gain4] Gain value for band with index 4
+   * @param {Number} [freq5] Frequency value for band with index 5
+   * @param {Number} [gain5] Gain value for band with index 5
+   * @param {Number} [freq6] Frequency value for band with index 6
+   * @param {Number} [gain6] Gain value for band with index 6
+   * @param {Number} [freq7] Frequency value for band with index 7
+   * @param {Number} [gain7] Gain value for band with index 7
```
therewasaguy: 16 parameters is a bit unwieldy, it seems like it would be hard to keep track of which value goes where

src/eq.js
```diff
@@ -0,0 +1,157 @@
...
+    //p5.EQ can be of size (3) or (8), defaults to 3
+    _eqsize = _eqsize === 3 || _eqsize === 8 ? _eqsize : 3;
+
+    var factor;
+    _eqsize == 3 ? factor = Math.pow(2,3) : factor = 2;
+
+    /**
+      *  The p5.EQ is built with abstracted p5.Filter objects. 
+      *  To modify any bands, use methods of the p5.Filter API.
+      *  To add or remove bands, use p5.EQ.addBand() and p5.EQ.removeBand().
```
therewasaguy: addBand and removeBand aren't implemented

src/eq.js
```diff
@@ -0,0 +1,157 @@
...
+  p5.EQ.prototype._newBand = function(freq, res) {
+    var newFilter = new Filter('peaking');
+    newFilter.disconnect();
+    newFilter.set(freq, res);
+    newFilter.biquad.gain.value = 0;
+    delete newFilter.input;
+    delete newFilter.output;
+    delete newFilter._drywet;
+    delete newFilter.wet;
+    return newFilter;
```
therewasaguy: I think we should overwrite the amp and dispose methods here if the filter should really have the same API. amp (and any other methods that we don't want/need) could console.warn a message to the developer that this method doesn't apply to EQ filters

src/eq.js
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  p5.EQ.prototype.dispose = function () {
+    Effect.prototype.dispose.apply(this);
+    
+    while (this.bands.length > 0) {
+      delete this.bands.pop().biquad.disconnect();
```
therewasaguy: nice workaround for the dispose method. I wonder if dispose method could be overwritten (above) to make it useable here in keeping with the Filter API. Maybe that's not necessary as the user wouldn't need to dispose of individual EQ filters...

therewasaguy:
@jvntf I'm working on a PR to your branch with a couple of the small tweaks

jvntf:
@therewasaguy cool great idea! i added a few more override methods to eqFilter

therewasaguy:
great work, @jvntf!

Issue #97, opened 2016-02-03
loadSound with a callback in preload never exits preload


therewasaguy:
https://github.com/processing/p5.js-sound/issues/88#issuecomment-179502439

This issue [was fixed in p5.js core library](https://github.com/processing/p5.js/pull/956) for methods that only have one callback method. Now that loadSound has success and error callbacks it is failing again.

PR #184, opened 2017-06-17
Logic to deal with callbacks in preload

bomoko:
This is an attempt at addressing issue #97. 

Some of the other `loadX()` functions in p5.js have logic to deal with the way `preload()` pushes its `decrementPreload()` function on the end of the argument list. 

This is causing some issues when `loadSound()` is called _with_ callbacks inside of preload (see the discussion [here](https://github.com/processing/p5.js/issues/2005))

I've tried to stick to the same kind of structure as some of the other `loadX()` functions.
What it does is fill in the arguments passed to `new p5.SoundFile()` in a way that's sensitive to the possibility that decrementPreload() has been passed as an argument. In the case that it is, it's called after the callback passed by the calling code (if any).

Any feedback on this would be really appreciated.

### DIFF
--- a/src/soundfile.js
+++ b/src/soundfile.js
@@ -177,7 +177,11 @@ define(function (require) {
    *  </code></div>
    */
   p5.prototype.loadSound = function(){
-    var callback, onerror, whileLoading;
+    var path = arguments[0]; 
+    var callback;
+    var onerror;
+    var whileLoading;
+
     var decrementPreload = p5._getDecrementPreload.apply(this, arguments);
       
     for(var i = 1; i < arguments.length; i++) {
@@ -198,9 +202,9 @@ define(function (require) {
       alert('This sketch may require a server to load external files. Please see http://bit.ly/1qcInwS');
     }
 
-    var s = new p5.SoundFile(arguments[0], function (soundFile) {
+    var s = new p5.SoundFile(path, function (sound) {
       if (typeof callback === 'function') {
-        callback(soundFile);
+        callback(sound);
       }
 
       if(decrementPreload) {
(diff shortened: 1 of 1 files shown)


src/soundfile.js
```diff
@@ -176,13 +176,37 @@ define(function (require) {
...
       alert('This sketch may require a server to load external files. Please see http://bit.ly/1qcInwS');
     }
 
-    var s = new p5.SoundFile(path, callback, onerror, whileLoading);
+    var s = new p5.SoundFile(arguments[0], function (soundFile) {
+      if (typeof callback === 'function') {
+        callback(soundFile);
+      }
+
+      if(decrementPreload) {
```
limzykenneth: `decrementPreload` will likely be defined as the `errorCallback` here, causing problem in https://github.com/processing/p5.js/issues/2007. Have a look at my PR https://github.com/processing/p5.js/pull/2008 but I'll add my workaround below as well. The rest of the changes are just as I would do it.

```
if(decrementPreload === onSuccess ||
   decrementPreload === onError){
   decrementPreload = null;
}
```

bomoko: Thanks @limzykenneth! I should've read the _getDecrementPreload() source before, I didn't realise it would return _any_ callback that's the last on the arguments list.
Let me make this change quick.

bomoko:
I'm going to close this for the moment. The issue that @limzykenneth pointed out is actually slightly more complex than I initially thought, and makes sense (I think) to discuss as an issue in the main p5.js repo.
Once that's cleared up (at least for me), I'll issue a new PR.

limzykenneth:
@bomoko Now that https://github.com/processing/p5.js/pull/2008 is merged, perhaps you can have a look at this again?

PR #185, opened 2017-06-18
add fft tests for smoothing and bins

therewasaguy:
- tweak how we initialize fft bins and smoothing (cc @TheBrenny)
- add some tests for p5.FFT


jvntf:
looks great to me!

PR #186, opened 2017-06-20
Linting: add grunt lint task, .eslintrc and .editorconfig

therewasaguy:
Adds linting, which is optional (for now) but hopefully helpful for developers who choose to use it. The goal is to have some consistent styles that nudge all contributors towards good standards/best practices without the need to get nitpicky in code review.

- Replaces the jshint task (which was similar to p5's) with an eslint task.
- Sets some linting rules in `src/.eslintrc`. Rules are based on [p5's jshint](https://github.com/processing/p5.js/blob/master/src/.jshintrc) and [jscs](https://github.com/processing/p5.js/blob/master/.jscsrc), with some additional suggestions. This is all open to input! Check out the list of available rules [here](http://eslint.org/docs/rules/).
- Adds an .editorconfig file [borrowed from p5 here](https://github.com/processing/p5.js/blob/master/.editorconfig)
- Lints source code to meet linting standards

I'm thinking we could add linting as a pre-commit hook, and then also set up linting on PR's, but I want to make sure this is done in a way that encourages, rather than deters, potential contributors. I would love some feedback on this idea!

##### Lint task
`npm install` to update dependencies
`grunt lint` lints `src/` directory

##### auto-fix errors
You can also run lint individual files from the terminal, with the optional -fix flag that will fix some (but not all) linting errors

##### Linting hints in your IDE
`npm install -g eslint`
`eslint src/audioin.js --fix`

You can also install tools for your code editor:
- eslint http://eslint.org/docs/user-guide/integrations
- editor config: http://editorconfig.org/#download


lmccart:
interesting! planning to take a look in the next few days

therewasaguy:
just pushed a commit that fixed some merge conflicts.

While I was at it I realized I might as well preserve some documentation for delay, distortion, and reverb methods that have been moved to p5.Effect like this: https://github.com/processing/p5.js-sound/pull/186/files#diff-361689f6101bd22d7f295e31cc8d4437R262 as discussed [here](https://github.com/processing/p5.js-sound/pull/178#pullrequestreview-45292854). @jvntf does this sound like a good approach?

lmccart:
looks good. do you find that the eslint gives better control than jshint? or was there another reason for the switch? wondering if we should consider this for p5.js, too.

therewasaguy:
@lmccart I don't know if it's worth p5 switching, but here are a few reasons I like eslint
- it provides both of what jshint and jscs offer with just one dependency
- one IDE plugin
- it is more flexible than jshint (but requires some more effort to set up)
- we're not taking advantage of this, but we can easily import eslint configs from other projects like [airbnb](https://www.npmjs.com/package/eslint-config-airbnb)...it'd be cool to have eslint-config-p5!
- we're not taking advantage of this yet but eslint has great es6+ support
- running `eslint --fix` can easily fix the more tedious things

PR #188, opened 2017-06-23
fixed drywet error with default values

jvntf:
found an error from Effect.js -- `drywet()` function throws an error when used as a getter


src/effect.js
```diff
@@ -87,7 +87,7 @@ define(function (require) {
 	 *	@param {Float}
 	 */
 	p5.Effect.prototype.drywet = function(fade){
-		this._drywet.fade.value = fade;
+		this._drywet.fade.value = fade || this._drywet.fade.value;
```
therewasaguy: if `fade` is zero, this technique treats `0` as "falsey". Another approach would be 

If zero should be allowed, then we really want to check for `typeof fade !== 'undefined'`

src/effect.js
```diff
@@ -87,7 +87,10 @@ define(function (require) {
 	 *	@param {Float}
 	 */
 	p5.Effect.prototype.drywet = function(fade){
-		this._drywet.fade.value = fade;
+
+		if (typeof fade !="undefined"){
```
therewasaguy: Best to use double equals to get the strict type checking 

`!== 'undefined'`

PR #192, opened 2017-06-26
Applying new loadXX _decrementPreload logic to loadSound

### LINKED ISSUE #193: Changes in the way p5.js deals with preload() breaks loadSound()
A change in the way p5.js deals with the `preload()` function (detailed in https://github.com/processing/p5.js/pull/2008) requires that any loadXX() function now needs to call `_decrementPreload()` directly.

Without calling `_decrementPreload()` directly, using `loadSound()` in `preload()` breaks the sketch (it stays in "loading" state forever).

bomoko:
Fixes #193.

This PR applies the logic in https://github.com/processing/p5.js/pull/2008 to the `loadSound()` function.

There is a full discussion of the issue in the main p5.js library's issue queue, including an example of how the current implementation breaks, at https://github.com/processing/p5.js/issues/2005#issuecomment-309017764.

### DIFF
--- a/src/soundfile.js
+++ b/src/soundfile.js
@@ -182,7 +182,15 @@ define(function (require) {
       alert('This sketch may require a server to load external files. Please see http://bit.ly/1qcInwS');
     }
 
-    var s = new p5.SoundFile(path, callback, onerror, whileLoading);
+    var self = this;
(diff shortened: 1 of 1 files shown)


therewasaguy:
thank you, @bomoko!

PR #199, opened 2017-07-08
yuidoc @type fixes

Spongman:
this fixes a few issues with the yuidoc @type declarations. mostly, `@type` doesn't use '{}' or take a description (even though the yuidoc docs suggest it does). also a few p5 namespace fixes.

### DIFF
--- a/src/delay.js
+++ b/src/delay.js
@@ -67,8 +67,7 @@ define(function (require) {
      *  <a href="http://www.w3.org/TR/webaudio/#DelayNode">
      *  Web Audio Delay Nodes</a>, one for each stereo channel.
      *
-     *  @property leftDelay
-     *  @type {Object}  Web Audio Delay Node
+     *  @property leftDelay {Object}  Web Audio Delay Node
      */
     this.leftDelay = this.ac.createDelay();
     /**
@@ -76,8 +75,7 @@ define(function (require) {
      *  <a href="http://www.w3.org/TR/webaudio/#DelayNode">
      *  Web Audio Delay Nodes</a>, one for each stereo channel.
      *
-     *  @property rightDelay
-     *  @type {Object}  Web Audio Delay Node
+     *  @property rightDelay {Object}  Web Audio Delay Node
      */
     this.rightDelay = this.ac.createDelay();
 
--- a/src/oscillator.js
+++ b/src/oscillator.js
@@ -181,7 +181,7 @@ define(function (require) {
    *  @param {Number} [rampTime] create a fade that lasts rampTime
    *  @param {Number} [timeFromNow] schedule this event to happen
    *                                seconds from now
-   *  @return  {AudioParam} gain  If no value is provided,
+   *  @return  {p5.AudioParam} gain  If no value is provided,
    *                              returns the Web Audio API
    *                              AudioParam that controls
    *                              this oscillator's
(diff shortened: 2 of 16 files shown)


therewasaguy:
Thanks @Spongman!

I'm getting an error when I try to view the p5.sound reference page after building the docs with these changes. Any idea what this could be about and is it working for you?

```
Uncaught TypeError: Cannot read property 'name' of undefined
    at n.eval (eval at m.template (underscore-min.js:5), <anonymous>:6:15)
    at n.c [as libraryTpl] (underscore-min.js:5)
    at n.render (reference.js:4422)
    at n.show (reference.js:4442)
    at reference.js:4713
    at reference.js:4607
    at Object.execCb (require.min.js:29)
    at $.check (require.min.js:18)
    at $.enable (require.min.js:23)
    at $.init (require.min.js:17)
```

therewasaguy:
...it seems p5.sound was appending/overwriting the p5.dom module. I'm not sure if that was related to these updates, but moving the below docblock to the `fragments/before.frag` did the trick

```
 *  @module p5.sound
 *  @submodule p5.sound
 *  @for p5.sound
 *  @main
```

Spongman:
hmmm. that's strange. i'm not seeing any error. i just did `grunt dev` and browsed around and i didnt' see any error. was there a specific page that was causing it?

src/delay.js
```diff
@@ -67,17 +67,15 @@ define(function (require) {
      *  <a href="http://www.w3.org/TR/webaudio/#DelayNode">
      *  Web Audio Delay Nodes</a>, one for each stereo channel.
      *
-     *  @property leftDelay
-     *  @type {Object}  Web Audio Delay Node
+     *  @property leftDelay {Object}  Web Audio Delay Node
```
therewasaguy: hmm, it seems I should get rid of the type descriptions because they wind up as part of the property name 

screenshot 2017-07-12 00 05 42

src/oscillator.js
```diff
@@ -497,8 +497,9 @@ define(function (require) {
    *  its method <code>setType('sine')</code>.
    *  See p5.Oscillator for methods.
    *
-   *  @method  p5.SinOsc
-   *  @param {[Number]} freq Set the frequency
+   *  @method  SinOsc
```
therewasaguy: these are actually constructors for classes that inherit from p5.Oscillator, and using `@ method` was just a hack to get them to appear on the p5.Oscillator page.

Do you happen to know a good way to ensure that these appear on the p5.Oscillator page, but not the p5 sound reference?

therewasaguy:
@Spongman the error happens on p5.sound reference page. `grunt dev` builds the library just fine, but when moving it over to the p5.js repo, running `grunt yui` there, and then copying data.min.json over to the p5.js-website repo and loading [the p5.sound reference page](https://p5js.org/reference/#/libraries/p5.sound) locally (whew!), that's where I was seeing the error

therewasaguy:
btw I started writing [this node script](https://gist.github.com/therewasaguy/8da82de2b948955d5baaa0fbffb71b0c) to make the above workflow of testing out changes to the docs a bit easier

src/delay.js
Spongman: i would definitely keep the property names. let's see if we can't fix the template so they don't appear where they're not wanted.

Spongman:
i'm trying to build the docs in upstream (without my changes) and I'm getting the following error:

````
 Warning: for p5.SoundFile.sampleRate() defined in lib/addons/p5.sound.js:1552, additional overloads should have no description ("<p>Return the sample rate of the sound file.</p>\n" !== "") Use --force to continue.
````

any ideas?

Spongman:
ok, i fetched upstream, temporarily removed the `@method sampleRate` attribute from `helpers.js` to fix the build problem, and i'm still seeing the `Cannot read property 'name' of undefined` error you mentioned initially, so i don't think that has anything to do with my changes.

therewasaguy:
yea I guess that issue is unrelated. I wonder how it arose but maybe it doesn't matter, I will fix it by moving the `@ module` block.

Thanks again for all your great work on this, @Spongman !

PR #203, opened 2017-07-15
edited documentation for compressor and effect

jvntf:


src/compressor.js
```diff
@@ -36,22 +33,33 @@ define(function (require) {
...
 	p5.Compressor.prototype.process = function(src, attack, knee, 
                                       ratio, threshold, release) {
 		src.connect(this.input);
 		this.set(attack, knee, ratio, threshold, release);
 	};
 
   /**
-   * Set the paramters of the compressor
-   *
+   * Set the paramters of the compressor.
```
therewasaguy: missing an "a"

src/compressor.js
```diff
@@ -65,7 +73,7 @@ define(function (require) {
 
 
   /**
-   * set attack w/ time ramp or get current attack
+   * Get current attack or set value w/ time ramp
    * @method attack
```
therewasaguy: Definitely include the descriptions you have above for `process` for all of these methods, they're really helpful.

Also, all of these methods should include info about the method's `@param`'s. Optional param names can be in square brackets like `[attack]`

`@param {Number} [attack] Time in seconds, between 0 and 1. Default 0.03.`

src/compressor.js
```diff
@@ -11,20 +11,17 @@ define(function (require) {
...
    * https://www.w3.org/TR/webaudio/#the-dynamicscompressornode-interface
    *
    * @class p5.Compressor
    * @extends p5.Effect
    * @constructor
    *
-   * @example
-   * <div><code>
-   *
-   * </code></div>
```
therewasaguy: looking forward to a Compressor example, maybe one that focuses on controlling threshold would be easiest to hear / visualize?

jvntf:
thanks for your reviews! -- 
re: `p5.Effect` methods, should these doc blocks be copied and pasted into classes that inherit? i think you have already done this in Delay for a couple of methods.

therewasaguy:
@jvntf good question. I don't think it'll be necessary to duplicate this documentation ourselves—we should be able to tell yuidoc to do it. So when we say a class `@extends {p5.Effect}`, yuidoc should be able to show p5.Effect methods documentation on the subclass. That doesn't seem to be enabled but it must be possible...@Spongman I'm curious if this is something you've looked into already?

Spongman:
Looking at the template code, I believe it should be possible to customize the rendered documentation to reference the base class. There's no need to duplicate the information in the source.

PR #205, opened 2017-07-19
graph routing error

jvntf:
found bug-- only left channel was getting connected under default Delay construction


therewasaguy:
nice, @jvntf !

PR #206, opened 2017-07-20
Simple presets

jvntf:


jvntf:
development on a preset library has been migrated to the [presets branch](https://github.com/processing/p5.js-sound/tree/presets)

PR #207, opened 2017-07-25
Update docs again

jvntf:
re-opening this PR because of some new fixes, and updated docs for all effects.

yuidoc tag @extends does not seem to render the relevant methods from the base class (does not seem to be happening for p5 either)-- have included links to base class / methods in the effect classes.

### DIFF
--- a/src/compressor.js
+++ b/src/compressor.js
@@ -90,6 +90,8 @@ define(function (require) {
 
   /**
    * Get current attack or set value w/ time ramp
+   * 
+   * 
    * @method attack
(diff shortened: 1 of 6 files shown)


therewasaguy:
I'd like to look into adding extends to p5's yuidoc, but this is great in the meantime!

therewasaguy:
here's a PR that'll automatically link to parent classes, though it doesn't add their methods to the page, or automatically link from parents to children https://github.com/processing/p5.js/pull/2098

PR #209, opened 2017-08-02
 p5.Looper WIP

jvntf:
A redesigned `p5.Looper`---
based on discussions with @therewasaguy re: a sort of draw loop for sound.
a simpler version of the `p5.Phrase / p5.Part / p5.Score` ecosystem
its just a loop! set an interval for how often a callback should be called.
use numerical notation to specify interval in seconds, or use [ToneJS time notation](https://github.com/Tonejs/Tone.js/wiki/Time#notation) which corresponds to the loopers BPM. Change the interval at any time and it will be reflected in the next iteration of the loop.

TODO: iteration counting / capping, clock syncing of multiple loops, triplets

`p5.Looper` is based on [Tone.Clock](https://tonejs.github.io/docs/#Clock) rather than [Tone.Transport](https://tonejs.github.io/docs/#Transport) because it is a relatively simple application, and I thought it wouldn't need all the features of Transport. also, to include less code in the library rather than more.


src/looper.js
```diff
@@ -1,520 +1,89 @@
...
-   */
-  p5.Part.prototype.getPhrase = function(name) {
-    for (var i in this.phrases) {
-      if (this.phrases[i].name === name) {
-        return this.phrases[i];
-      }
-    }
+    this.interations;
+    this.clock = new Clock({
+      "callback" : this.callback,
```
therewasaguy: Should users be able to access / modify the callback on the fly? Maybe that's not necessary, but it is misleading to be able to access `this.callback` because changing it does nothing.

src/looper.js
```diff
@@ -1,520 +1,89 @@
...
-   *  @method  getPhrase
-   *  @param  {String} phraseName
-   */
-  p5.Part.prototype.getPhrase = function(name) {
-    for (var i in this.phrases) {
-      if (this.phrases[i].name === name) {
-        return this.phrases[i];
-      }
-    }
+    this.interations;
```
therewasaguy: Should this be `iterations`? It doesn't appear to be in use just yet—let's keep any properties out of the PR until they become necessary to stay focused.

src/looper.js
```diff
@@ -1,520 +1,89 @@
...
-   *  @param  {String} phraseName
-   */
-  p5.Part.prototype.removePhrase = function(name) {
-    for (var i in this.phrases) {
-      if (this.phrases[i].name === name) {
-        this.phrases.splice(i, 1);
-      }
-    }
-  };
+    this.bpm = BPM || 60;
```
therewasaguy: users need to be able to modify the BPM on the fly, I think it's a bit misleading that users could access `this.bpm` and even change it, but that wouldn't impact the tick frequency. Let's figure out a way to do this. In the current implementation, we'd need to set `myLooper.clock.frequency.value` (but convert bpm to frequency);

src/looper.js
therewasaguy: I think we may want/need to update the callback to bind its scope (`this` within the context of the callback) to the p5.Looper, rather than to the Click. We can do this by creating a reference to this instance of the looper class (the variable name `self` is a convention but it could be anything), and then calling `self.callback`

We may also need to do this in order to pass in "time from now" to the callback, because that is what all existing p5.sound functions that accept a time are expecting. Tone follows the Web Audio convention of passing in time from the beginning of the web audio context.

```
    var self = this;
    this.clock = new Clock({
      "callback" : function(time) {
        var timeFromNow = time - p5sound.audiocontext.currentTime;
        self.callback(timeFromNow);
      },
```

jvntf:
made most of these changes! 
regarding `.setBPM` and `.setTimeSignature` ---
I am thinking that perhaps the looper should have two _modes_. For example, if the interval is specified in seconds, changing the BPM and / or timesignature will affect the length of the loop (currently). If the loop is 8 seconds at 4/4 and 60BPM, it will become 4 seconds at 4/4 and 120BPM... this change in the loop duration makes sense when the duration is specified as "4n" or "1m"--- 
what do you think? @therewasaguy

jvntf:
closing this pull request, opening new one for ease of viewing changes. see [#210](https://github.com/processing/p5.js-sound/pull/210)

PR #210, opened 2017-08-04
p5.SoundLoop()

jvntf:
A redesigned `p5.Looper`---
based on discussions with @therewasaguy re: a sort of draw loop for sound.
a simpler version of the `p5.Phrase / p5.Part / p5.Score` ecosystem
its just a loop! set an interval for how often a callback should be called.
use numerical notation to specify interval in seconds, or use [ToneJS time notation](https://github.com/Tonejs/Tone.js/wiki/Time#notation) which corresponds to the loopers BPM. Change the interval at any time and it will be reflected in the next iteration of the loop.

TODO: iteration counting / capping, clock syncing of multiple loops, triplets

`p5.Looper` is based on [Tone.Clock](https://tonejs.github.io/docs/#Clock) rather than [Tone.Transport](https://tonejs.github.io/docs/#Transport) because it is a relatively simple application, and I thought it wouldn't need all the features of Transport. also, to include less code in the library rather than more.


examples/sound_loop/sketch.js
```diff
@@ -0,0 +1,54 @@
...
+	count2 = 0;
+	c = color(50);
+	d = color(50);
+  // create a part with 8 spaces, where each space represents 1/16th note (default)
+  looper1 = new p5.SoundLoop(function(time){
+    click.play();
+    console.log('looper1 '+this.clock._nextTick);
+    }, 1);
+
+  looper2 = new p5.SoundLoop(function(time){
```
therewasaguy: maybe we should change this to be `timeFromNow` to be clearer ?

examples/sound_loop/sketch.js
```diff
@@ -0,0 +1,54 @@
...
+    
+	beatbox.play();
```
therewasaguy: we should pass `timeFromNow` in as a parameter

examples/sound_loop/sketch.js
```diff
@@ -0,0 +1,54 @@
...
+
+function setup() {
+	createCanvas(500,500)
+	count1 = 0;
+	count2 = 0;
+	c = color(50);
+	d = color(50);
+  // create a part with 8 spaces, where each space represents 1/16th note (default)
+  looper1 = new p5.SoundLoop(function(time){
+    click.play();
```
therewasaguy: pass in `timeFromNow`

src/soundLoop.js
```diff
@@ -0,0 +1,228 @@
...
+    //8 second interval at 120BPM in 5/4 time will be 3.2 seconds long
+    this._timeSignature = 4;
+    this._bpm = BPM || 60;
+
+    this.isPlaying = false;
+    var self = this;
+    this.clock = new Clock({
+      'callback' : function(time) {
+        var timeFromNow = time - p5sound.audiocontext.currentTime;
+        if (timeFromNow > 0) {self.callback(timeFromNow);}
```
therewasaguy: why would this be less than zero anyway? ah, @jvntf and I are testing together and found that this is the case when we load the page. Maybe worth adding a comment?

src/soundLoop.js
```diff
@@ -0,0 +1,228 @@
...
+  /**
+   * Calculate the frequency of the clock's callback based on bpm, interval, and timesignature
+   * @private
+   * @method  _calcFreq
+   * @return {Number} new clock frequency value
+   */
+  p5.SoundLoop.prototype._calcFreq = function() {
+    if (typeof this._interval === 'number') {
+      this.musicalTimeMode = false;
+      return this._bpm / 60 / this._interval * (this._timeSignature / 4);
```
therewasaguy: it seems like this should just set the interval in seconds

PR #214, opened 2017-08-16
Synths

jvntf:
built on the work of @b2renger in the [synths](https://github.com/processing/p5.js-sound/tree/synths) branch--

follows the original design...
adds an AudioVoice class, MonoSynth class that extends AudioVoice, and PolySynth class as a manager 

PolySynth class is redesigned to use resources better, follow polyphony rules, etc

opened PR with `master` as opposed to `synths` because the synths branch is quite old and the amount of changes that show up in the diff make it confusing


src/oscillator.js
```diff
@@ -236,11 +236,12 @@ define(function (require) {
       var now = p5sound.audiocontext.currentTime;
       var rampTime = rampTime || 0;
       var tFromNow = tFromNow || 0;
+      var t = now + tFromNow + rampTime;
       // var currentFreq = this.oscillator.frequency.value;
       // this.oscillator.frequency.cancelScheduledValues(now);
 
       if (rampTime === 0) {
-        this.oscillator.frequency.cancelScheduledValues(now);
+        // this.oscillator.frequency.cancelScheduledValues(now);
```
jvntf: commenting out this line solves an error with scheduling the changes of values. i think that we _don't_ want to 'cancelScheduledValues' because, doing results in losing every note except the last one scheduled.
ex. if we have a monosynth and
'monosynth.play(60,1,0,1);
monosynth.play(64,1,1,1);
monosynth.play(68,1,2,1);'
we will hear note 69 (A = 440) twice and note 68 (Ab = 415.3), because each value cancels all preceding values.

src/monosynth.js
```diff
@@ -0,0 +1,358 @@
...
+     *  For instance if you want to build a complex synthesizer
+     *  with one or more filters, effects etc. this is where you
+     *  will want to set their values.
+     *  
+     *  @method  setParams
+     *  @param   
+     * 
+     */  
+
+  p5.MonoSynth.prototype.setParams = function(params) {
```
therewasaguy: let's leave this method out since it does nothing, it'll be distracting for most users. We could instead include an example for how to extend the MonoSynth

src/monosynth.js
```diff
@@ -0,0 +1,358 @@
...
+     *  @method playNote 
+     *  @param  {Number} [note] midi note to play (ranging from 0 to 127 - 60 being a middle C)
+     *  @param  {Number} [velocity] velocity of the note to play (ranging from 0 to 1)
+     *  @param  {Number} [secondsFromNow]  time from now (in seconds) at which to play
+     *  @param  {Number} [sustainTime] time to sustain before releasing the envelope
+     *  
+     */  
+
+  p5.MonoSynth.prototype.play = function (note, velocity, secondsFromNow, susTime) {
+    // set range of env (TO DO: allow this to be scheduled in advance)
```
therewasaguy: yes! the env `ramp` has a `secondsFromNow` method

src/oscillator.js
therewasaguy: ah, nice find

src/monosynth.js
```diff
@@ -0,0 +1,358 @@
...
+    this.setADSR(0.02, 0.25, 0.05, 0.35);
+
+    // filter
+    this.filter = new p5.Filter('highpass');
+    this.filter.set(5, 1);
+
+    // oscillator --> env --> filter --> this.output (gain) --> p5.soundOut
+
+    this.oscillator.connect(this.filter);
+    this.env.setInput(this.oscillator);
```
therewasaguy: I'm not sure if the chain is working (testing it out, all the examples are silent)

Maybe we just need `p5.MonoSynth.prototype = Object.create(p5.AudioVoice.prototype);`

src/monosynth.js
```diff
@@ -0,0 +1,358 @@
...
+     *  @param  {Number} [secondsFromNow]  time from now (in seconds) at which to play
+     *  @param  {Number} [sustainTime] time to sustain before releasing the envelope
+     *  
+     */  
+
+  p5.MonoSynth.prototype.play = function (note, velocity, secondsFromNow, susTime) {
+    // set range of env (TO DO: allow this to be scheduled in advance)
+    var vel = velocity || 1;
+
+    this.triggerAttack(note,velocity,secondsFromNow);
```
therewasaguy: should this be `vel` ?

src/monosynth.js
```diff
@@ -0,0 +1,358 @@
...
+  p5.MonoSynth.prototype.triggerAttack = function (note, velocity, secondsFromNow) {
+
+    var now = p5sound.audiocontext.currentTime;
+    var tFromNow = secondsFromNow || 0;
+    var t = now + tFromNow;
+    var n = p5.prototype.midiToFreq(note);
+    
+    this._isOn = true;
+    this.oscillator.freq(n, 0, t);
+    this.env.ramp(this.output, t, velocity);
```
therewasaguy: I think these values should be `secondsFromNow`, because they will add `audiocontext.currentTime` - so don't add `now` just yet

src/monosynth.js
```diff
@@ -0,0 +1,358 @@
...
+     *  @param  {Number} [sustainTime] time to sustain before releasing the envelope
+     *  
+     */  
+
+  p5.MonoSynth.prototype.play = function (note, velocity, secondsFromNow, susTime) {
+    // set range of env (TO DO: allow this to be scheduled in advance)
+    var vel = velocity || 1;
+
+    this.triggerAttack(note,velocity,secondsFromNow);
+    this.triggerRelease(secondsFromNow + susTime);
```
therewasaguy: I think we need a `susTime`?

src/monosynth.js
```diff
@@ -0,0 +1,358 @@
...
+
+
+  /**
+   * Getters and Setters
+   * @param {Number} attack
+   * @param {Number} decay
+   * @param {Number} sustain
+   * @param {Number} release
+   */
+  Object.defineProperties(p5.MonoSynth, {
```
therewasaguy: define these on the `prototype`, so that they're properties of monosynth instances

examples/monosynth_basic/sketch.js
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+var monoSynth;
+
+function setup() {
+  monoSynth = new p5.MonoSynth();
+}
+
+function keyPressed() {
+  // pick a random midi note
```
therewasaguy: maybe we should use mouse because it'll then be accessible on touch devices, and because keyPressed will fire more than once

via https://p5js.org/reference/#/p5/keyPressed:

> Because of how operating systems handle key repeats, holding down a key may cause multiple calls to keyTyped() (and keyReleased() as well). The rate of repeat is set by the operating system and how each computer is configured.

src/monosynth.js
jvntf: this is why no sound was coming out...i over scheduled. thanks for the catch!

jvntf:
updated this branch to include commits from the old [synths branch](https://github.com/processing/p5.js-sound/tree/synths)

therewasaguy:
This is looking pretty good to me! @b2renger curious if you have any thoughts as this builds on what you started? 

I just sent a PR (https://github.com/jvntf/p5.js-sound/pull/4) that fixes an issue I had when running locally that you may not have had, @jvntf. 

I want to see what the generated reference pages will look like to users, but I'm getting an error when compiling documentation (the error is unrelated to this PR so I'll look at that separate). And maybe we should tackle documentation separately.

We'll also want to clean up the examples related to synths in the `/examples` directory to distinguish between examples that use synths from the library, and examples that build their own (which I think are good to have as well?)

b2renger:
@jvntf this is really cool :) @therewasaguy  I'll try to look it up soon (today or tomorrow) and I'll try to look at the spatializer branch that was merged as well (sometime later on)

src/monosynth.js
```diff
@@ -0,0 +1,268 @@
...
+     *  @param  {Number} [secondsFromNow]  time from now (in seconds) at which to play
+     *  @param  {Number} [sustainTime] time to sustain before releasing the envelope
+     *
+     */
+
+  p5.MonoSynth.prototype.play = function (note, velocity, secondsFromNow, susTime) {
+    // set range of env (TO DO: allow this to be scheduled in advance)
+    var susTime = susTime || this.sustain;
+    this.susTime = susTime;
+    this.triggerAttack(note, veocity, secondsFromNow);
```
therewasaguy: `veocity` --> `velocity`

Running eslint should catch things like this

src/monosynth.js
```diff
@@ -0,0 +1,268 @@
...
+     *  hold the sustain level until you let go.
+     *
+     *  @param  {Number} secondsFromNow time from now (in seconds)
+     *  @param  {Number} [velocity] velocity of the note to play (ranging from 0 to 1)
+     *  @param  {Number} [secondsFromNow]  time from now (in seconds) at which to play
+     *  @method  triggerAttack
+     */
+  p5.MonoSynth.prototype.triggerAttack = function (note, velocity, secondsFromNow) {
+    var secondsFromNow = secondsFromNow || 0;
+    var freq = p5.prototype.midiToFreq(note);
```
therewasaguy: I wonder if these methods should use frequency, and let users decide whether they want to work in midi by utilizing `midiToFreq` in the synth examples?

jvntf: hm... i agree that playing frequency is important, but i like the ease of playing midi notes. maybe similar format to the soundLoop of [two modes](https://github.com/processing/p5.js-sound/blob/master/src/soundLoop.js#L260) differentiated by String and Number would be good. Frequency = Number, midi notes ("C4", "Eb3"...etc) = String ?

jvntf:
@therewasaguy re: the `\examples` directory I am thinking 
`\examples\monosynth_basic\` is a nice basic example that should stay as is
`\examples\polyphonic_synth\` is a very cool example that should stay as an example of how to extend and build your own synths (needs some updating due to API change)
`\examples\polysynth_old\` is also cool but seems like it maybe can just be combined with `\examples\polyphonic_synth\`? (this example contains a few more extensions of `AudioVoice class` (needs some updating due to API change)
`\examples\polyphonicSynth-Keyboard\` is a 15-key keyboard example using `p5.PolySynth` and `p5.MonoSynth`
i'm happy to do these updates / migrations

src/monosynth.js
b2renger: Using midiToFreq is good but the two modes is probably the better option since it is used in the soundLoop class (that I have to check !) and allows both

b2renger:
@jvntf @therewasaguy 

All your work is great on this subject. I think it's ready to merge when the examples are done. This is awesome.

As for the example I agree with you @jvntf  the wrappers from the psynth_old should be merged in the new example and maybe we could switch beetween extensions with keyboard interactions ?

About the keyboard example it sounds a bit weird when several keys are pressed together, is it because of the synth ? because it sound like we cannot play a simple 2-note or 3-note chord.

Lastly the granular_sampler_psynth can probably be deleted from the master branch, those classes don't have to deal with sample stuff.

b2renger:
@jvntf I was thinking about the example with the keyboard and the fact that everything was distorted when playing more than one voice. 

And I remembered I had tried to reduce the env range according to the number of voices used in a polysynth like this : https://github.com/processing/p5.js-sound/pull/214/commits/472147017f16aa3e0298581c84a1b33edc3e116d

maybe it could be a good idea to have this kind of mechanism here too

jvntf:
@b2renger when you say distorted, are you referring to the clipping that sometimes happens? I just added in a maxRange for the envelope per your suggestion! let me know what you think-- to me it seems that it sounds pretty good considering the current p5.MonoSynth is fairly minimal in terms of sound design

therewasaguy:
@jvntf I think this distortion results from playing multiple notes with an amplitude envelope that maxes out at 1.0. Especially when the notes are in a similar frequency range, some speakers just can't handle it.

@b2renger I like the approach of setting the envelope's attack range based on the number of voices.

Building on that idea, @jvntf has set us up to track `_voicesInUse`, and we could use that to modulate the polysynth's output amplitude based on the number of active voices. This way, if just one note is active, it can still be loud—we only need to turn it down once other voices join in.

I think we should go ahead and merge this and address the distortion in a separate PR. I was thinking to squash some of these commits down because there are so many, but it's tricky, especially with so many contributors, and I think the commit history can be valuable like the way @b2renger referenced https://github.com/processing/p5.js-sound/commit/472147017f16aa3e0298581c84a1b33edc3e116d. So let's just merge it with all the history.

Very excited about all of this!

PR #215, opened 2017-08-20
Preset library

jvntf:
migrating work on presets to a new branch.
Preset library is up for discussion re: how to store and load presets into effects and synth instruments

currently includes synths #214 work tree as that PR is unmerged


therewasaguy:
awesome, I'm excited about this branch!

PR #217, opened 2017-08-20
update docs for Panner3d

therewasaguy:
- fix duplicate method yuidoc definitions
- update p5.Panner3D description

@jvntf this fixes an issue when running `grunt yui` to generate the docs in p5.js repo

`
Warning: for Spatializer.positionZ() defined in lib/addons/p5.sound.js:8442, additional overloads should have no description ("<p>Getter and setter methods for orient coordinates</p>\n" !== "") Use --force to continue.
`

...but I'm not sure of the best way to get sub-properties of the p5.Panner3d.panner (distanceModel, panningModel) to display properly in our reference pages.

### DIFF
--- a/src/panner3d.js
+++ b/src/panner3d.js
@@ -9,7 +9,7 @@ define(function (require) {
    * "https://developer.mozilla.org/en-US/docs/Web/API/PannerNode">
    * Web Audio Spatial Panner Node</a>.
    * This panner is a spatial processing node that allows audio to be positioned
-   * and oriented in 3D space. 
+   * and oriented in 3D space.
    *
    * The position is relative to an <a title="Web Audio Listener docs" href=
    * "https://developer.mozilla.org/en-US/docs/Web/API/AudioListener">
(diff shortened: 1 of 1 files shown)


src/panner3d.js
```diff
@@ -5,29 +5,36 @@ define(function (require) {
...
 	p5.Panner3D = function() {
       Effect.call(this);
+
+      /**
+       *  <a title="Web Audio Panner docs"  href=
+       *  "https://developer.mozilla.org/en-US/docs/Web/API/PannerNode">
+       *  Web Audio Spatial Panner Node</a>
+       *
+       *  Properties include
+       *    -  panningModel: "equal power" or "HRTF"
```
jvntf: I think this is ok, maybe we should just add links to the Web Audio docs for these properties? [distanceModel](https://www.w3.org/TR/webaudio/#idl-def-DistanceModelType) and [panningModel](https://www.w3.org/TR/webaudio/#idl-def-PanningModelType)

therewasaguy: great idea, just added links

PR #219, opened 2017-10-02
Fix comment in p5.Oscillator reference example

dcbriccetti:
Fix comment in p5.Oscillator reference example, https://p5js.org/reference/#/p5.Oscillator.


therewasaguy:
thank you, @dcbriccetti !

dcbriccetti:
Glad to help a project so cool.

PR #223, opened 2017-10-15
Added Playmode 'untilDone'

pjryan93:
Fix for https://github.com/processing/p5.js-sound/issues/5. I thought this would be the cleanest way to implement this, the return statement might be a little weird.


therewasaguy:
🎉 thanks @pjryan93 !

PR #230, opened 2017-11-30
fix documentation bug

Spongman:
when the combined `data.json` file is generated, the source file for the `p5` class was listed as `addons/p5.sound.js`. this is because of this bogus `@class p5` directive in `src/helpers.j`.

this PR switches this to a `@for` directive which indicated that these are additional methods for a class defined elsewhere...


therewasaguy:
thank you, @Spongman !

PR #232, opened 2017-12-09
fixed broken link in getEnergy() example

elizasj:
I fixed a broken link the getEnergy() example html, which figures as commented out text in the codebase, line 302. All the other mods bellow are just from my IDE which is set up w/Prettier which swapped out some double quotes for singles (sorry about that!) 

That broken link has been bothering me for ages, hopefully you don't mind that I fixed such a small thing!

### DIFF
--- a/src/fft.js
+++ b/src/fft.js
@@ -1,6 +1,6 @@
 'use strict';
 
-define(function (require) {
+define(function(require) {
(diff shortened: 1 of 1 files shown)


therewasaguy:
thank you, @elizasj ! 👍 to the code formatting, too!

PR #238, opened 2018-01-12
Fix missing paren on p5.SoundLoop example sketch

DexterShepherd:
Hey all,

Noticed we were missing a paren after the loadSound call on the p5.SoundLoop example sketch. This was breaking the example and the formatting on the p5 reference site.

All the best

Dex


therewasaguy:
thank you @Dex!

PR #248, opened 2018-01-29
remove extraneous docs from overridden Noise methods.

Spongman:
these docs in `noise.js` are duplicate docs for methods defined in the base class. i have removed them from the derived class because they are out of sync with the base definitions and are causing errors with the typescript definition build.

### DIFF
--- a/src/noise.js
+++ b/src/noise.js
@@ -112,11 +112,6 @@ define(function (require) {
     return this.buffer.type;
   };
 
-  /**
-   *  Start the noise
-   *
-   *  @method start
-   */
   p5.Noise.prototype.start = function() {
     if (this.started) {
       this.stop();
@@ -130,11 +125,6 @@ define(function (require) {
     this.started = true;
   };
 
-  /**
-   *  Stop the noise.
-   *
-   *  @method  stop
-   */
   p5.Noise.prototype.stop = function() {
     var now = p5sound.audiocontext.currentTime;
     if (this.noise) {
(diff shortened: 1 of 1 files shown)


therewasaguy:
thank you @Spongman!

Spongman:
Thank you!

JTBrunderley:
Hi @therewasaguy and @Spongman! I have run into this issue while using the p5.js 0.6.0 release from inside an angular5 js. I understand I can resolve this issue manually by correcting the return type in the typescript definition file, so working with it locally is no issue. However, when I try to deploy my app to Heroku, it downloads the files through node during deployment and I don't have the ability to fix it there, so my deployments all fail.  I see this was merged into master a few months ago, do you know when the version containing the fix will be released and available through node?

Thank you for the help and your work here!

JTBrunderley:
**For anyone looking for a work-around until until this fix is included in an official p5 release:**

I created a windows batch file to manually correct the Noise amp method's return type in the p5.d.ts file.

Current Incorrect Line:  amp(volume: number|object, rampTime?: number, timeFromNow?: number): void

Corrected Line:  amp(volume: number|object, rampTime?: number, timeFromNow?: number): AudioParam

Inside file_fix.bat:

@echo off &setlocal
setlocal enabledelayedexpansion

set "search=    amp(volume: number|object, rampTime?: number, timeFromNow?: number): void"
set "replace=    amp(volume: number|object, rampTime?: number, timeFromNow?: number): AudioParam"
set "textfile=%~dp0node_modules\p5\lib\p5.d.ts"
set "newfile=%~dp0node_modules\p5\lib\p5_1.d.ts"
(for /f "delims=" %%i in (%textfile%) do (
    set "line=%%i"
    set "line=!line:%search%=%replace%!"
    echo(!line!
))>"%newfile%"
del %textfile%
rename %newfile%  p5.d.ts
endlocal

I run this batch file from node's package.json postinstall script. This way the p5.d.ts file is corrected directly after it is installed:

 "postinstall": ".\\file_fix.bat"

I have not tried this piece yet, but if you are deploying to Heroku, I believe you should be able to set your postInstall like this:

 "postinstall": ".\\file_fix.bat && ng build --aot -prod"

This should fix the file before Heroku starts building the app.

JTBrunderley:
The above will not work for heroku since it apparently runs on Linux. You'll need to create a sh script instead of using the windows batch script above.

PR #256, opened 2018-02-25
fix getCentroid reference page alignment

emclaren:
This pull request resolves the issue listed in processing/p5.js-website#160 & processing/p5.js#2591

In the [reference docs](https://p5js.org/reference/#/p5.FFT/getCentroid), the canvas overlaps with the example code. 

Fixed by reducing the size of the canvas to the default size of 100x100. I made the text smaller so it would fit properly on the small canvas.

I hope this is the right place to make this update.

### DIFF
--- a/src/fft.js
+++ b/src/fft.js
@@ -401,7 +401,7 @@ define(function(require) {
    *
    *
    *function setup(){
-   *  cnv = createCanvas(800,400);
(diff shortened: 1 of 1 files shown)


therewasaguy:
thank you, @emclaren !

PR #270, opened 2018-04-04
Preload fixes

Spongman:
this PR fixes a few issues found when running the p5.js reference tests in browser:
- ensures `loadSound()` is called from `preload`. this prevents the p5 instance from being removed before the load has completed which can cause issues especially during the tests when multiple p5 instances are created one after the other.
- ensures that `createConvolver()` calls `_decrementPreload`. this allows preload functions that create convolvers to complete.
- fixes some asset paths.

### DIFF
--- a/src/reverb.js
+++ b/src/reverb.js
@@ -316,7 +316,13 @@ define(function (require) {
     if (window.location.origin.indexOf('file://') > -1 && window.cordova === 'undefined') {
       alert('This sketch may require a server to load external files. Please see http://bit.ly/1qcInwS');
     }
-    var cReverb = new p5.Convolver(path, callback, errorCallback);
+    var self = this;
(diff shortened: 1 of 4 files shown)


src/reverb.js
```diff
@@ -316,7 +316,13 @@ define(function (require) {
     if (window.location.origin.indexOf('file://') > -1 && window.cordova === 'undefined') {
       alert('This sketch may require a server to load external files. Please see http://bit.ly/1qcInwS');
     }
-    var cReverb = new p5.Convolver(path, callback, errorCallback);
+    var self = this;
+    var cReverb = new p5.Convolver(path, function(buffer) {
+      if (typeof callback === 'function') {
+        callback(buffer);
+      }
+      self._decrementPreload();
```
therewasaguy: good call, thanks @Spongman !

PR #275, opened 2018-05-05
protect against errors during duplicate dispose/disconnect calls

Spongman:
i have added some extra protection around the dispose/disconnect calls. in normal use these errors don't usually occur, but in the context of the p5 tests where many p5 instances are continuously brought up and torn down, sometimes these trigger.

this is in order to get the p5.js tests to all pass. specifically, we're investigating using headless-chrome instead of phantomjs which allows the webl & sound tests to run as part of the regular travis build-validation pass.

### DIFF
--- a/src/soundfile.js
+++ b/src/soundfile.js
@@ -328,6 +328,11 @@ define(function (require) {
    * @param {Number} [duration]          (optional) duration of playback in seconds
    */
   p5.SoundFile.prototype.play = function(startTime, rate, amp, _cueStart, duration) {
+    if (!this.output) {
+      console.warn('SoundFile.play() called after dispose');
+      return;
+    }
+
     var self = this;
     var now = p5sound.audiocontext.currentTime;
     var cueStart, cueEnd;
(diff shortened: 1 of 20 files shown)


src/soundfile.js
```diff
@@ -328,6 +328,11 @@ define(function (require) {
    * @param {Number} [duration]          (optional) duration of playback in seconds
    */
   p5.SoundFile.prototype.play = function(startTime, rate, amp, _cueStart, duration) {
+    if (!this.output) {
+      console.warn('SoundFile.play() called after dispose');
+      return;
+    }
```
therewasaguy: I'm wondering why put a warning here, and not other places where a method might be called post-dispose? I'd rather not add complexity to the play method because it's already more complex than it needs to be.

Spongman: this was just the one method that i came across during the tests where user code called a library function after the library objects had been disposed (i believe it was something to do with a loop timer).

the other places I changed were just errors caused during p5 instance teardown, so i figured a warning wasn't really appropriate since they're much more likely to happen and they're not really due to user error.

PR #278, opened 2018-05-11
Fix effect class extension reference urls

koltclassic:
This PR updates the class extension reference urls in p5 Effect as they are currently 404-ing, as well as some minor whitespace corrections.


koltclassic:
Oh, sorry about that! Thank you for the information!

PR #279, opened 2018-05-18
Add new example to demonstrate envelope ADSR shape

JunShern:
When I was trying to understand the ADSR parameters, I felt it might be helpful to have something like this: Use sliders to control parameters of envelope, with plotted shape to illustrate changes and click to play sound. 

Changes:
- New `envelope_designer` in `/examples` folder, with corresponding `index.html` and `sketch.js`
- Updated `/lib/addons/p5.dom.js` with latest version from website (needed latest code to allow continuous-valued sliders)

Feedback welcome! :smiley:

### DIFF
--- a/examples/envelope_designer/sketch.js
+++ b/examples/envelope_designer/sketch.js
@@ -1,0 +1,103 @@
+var attackTime = 0.001
+var attackLevel = 1.0;
+var decayTime = 0.2;
+var decayLevel = 0.2;
+var releaseTime = 0.5;
+var releaseLevel = 0;
+var sustainTime = 0.5;
+
+var attackTimeInput, attackLevelInput;
+var decayTimeInput, decayLevelInput;
+var releaseTimeInput, releaseLevelInput;
+var sustainTimeInput;
+
+var maxTime = 1;
+var maxLevel = 5;
+var xScale = 100;
+
+var env, osc;
+
+function setup() {
+  var cnv = createCanvas(500, 500);
+  cnv.mousePressed(playEnv);
+
+  env = new p5.Env();  
+  osc = new p5.Oscillator('triangle');
+  osc.amp(env);
+  osc.start();
+  osc.freq(440);
+
+  attackTimeInput = createSlider(0, maxTime, attackTime, 0);
+  attackTimeInput.position(20, 15);
+  attackLevelInput = createSlider(0, maxLevel, attackLevel, 0);
+  attackLevelInput.position(160, 15);
+  decayTimeInput = createSlider(0, maxTime, decayTime, 0);
+  decayTimeInput.position(20, 45);
+  decayLevelInput = createSlider(0, maxLevel, decayLevel, 0);
+  decayLevelInput.position(160, 45);
+  sustainTimeInput = createSlider(0, maxTime, sustainTime, 0);
+  sustainTimeInput.position(20, 75);
+  releaseTimeInput = createSlider(0, maxTime, releaseTime, 0);
+  releaseTimeInput.position(20, 105);
+  releaseLevelInput = createSlider(0, maxLevel, releaseLevel, 0);
+  releaseLevelInput.position(160, 105);
+}
+
+function draw() {
+  background(30, 100, 155);
+  attackTime = attackTimeInput.value();
+  attackLevel = attackLevelInput.value();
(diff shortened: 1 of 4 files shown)


examples/envelope_designer/sketch.js
```diff
@@ -0,0 +1,103 @@
+var attackTime = 0.001
```
therewasaguy: I wonder if we should lower the range of the slider, or think of some other interface (perhaps a number input), since a value of 0.1 vs 0.001 second makes an audible difference? On the other hand, that would make the code more complex...

examples/envelope_designer/sketch.js
```diff
@@ -0,0 +1,103 @@
...
+  fill(130, 200, 255);
+  noStroke();
+  textAlign(LEFT, BOTTOM);
+
+  beginShape();
+  // Start
+  vertex(0, height);
+  // Attack
+  x = attackTime*xScale;
+  y = height - attackLevel*yScale;
```
therewasaguy: x and y are global variables because they weren't declared with `var`, 

I think it make sense to make them local to the drawADSR method. And since we keep redefining them, we can declare them (without instantiating them) at the beginning of the drawADSR method along with the other local variables.

examples/envelope_designer/sketch.js
```diff
@@ -0,0 +1,103 @@
...
+  decayTimeInput = createSlider(0, maxTime, decayTime, 0);
+  decayTimeInput.position(20, 45);
+  decayLevelInput = createSlider(0, maxLevel, decayLevel, 0);
+  decayLevelInput.position(160, 45);
+  sustainTimeInput = createSlider(0, maxTime, sustainTime, 0);
+  sustainTimeInput.position(20, 75);
+  releaseTimeInput = createSlider(0, maxTime, releaseTime, 0);
+  releaseTimeInput.position(20, 105);
+  releaseLevelInput = createSlider(0, maxLevel, releaseLevel, 0);
+  releaseLevelInput.position(160, 105);
```
therewasaguy: I think these UI elements need some labels, as it wasn't clear to me which slider controlled which parameter. We don't have a label element in p5.dom, but we should! https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

examples/envelope_designer/sketch.js
```diff
@@ -0,0 +1,103 @@
...
+  drawADSR();
+
+  // Text
+  textAlign(CENTER);
+  text('click to play', width/2, height/2);
+}
+
+function playEnv() {
+  env.set(attackTime, attackLevel, decayTime, decayLevel, releaseTime, releaseLevel);
+  env.play(osc, 0, sustainTime);
```
therewasaguy: A bit of a tangent, but I thought it might be interesting to see these methods as part of the example.

And then, I thought of this classic essay by Bret Victor
http://worrydream.com/LearnableProgramming/

(which may have inspired [Processing Tweak Mode](http://galsasson.com/tweakmode/))

...could be good inspiration as you create more examples and think about the user experience from both a code and UI perspective.

examples/envelope_designer/sketch.js
```diff
@@ -0,0 +1,103 @@
...
+  textAlign(LEFT, BOTTOM);
+
+  beginShape();
+  // Start
+  vertex(0, height);
+  // Attack
+  x = attackTime*xScale;
+  y = height - attackLevel*yScale;
+  vertex(x, y);
+  text(`Attack: \n${attackTime.toFixed(3)}s \n${attackLevel.toFixed(3)}`, x, y);
```
therewasaguy: these labels and line breaks can get a bit cluttered in the bottom left corner. Maybe they just need one label?

examples/envelope_designer/sketch.js
JunShern: Oops! Yep, this was accidental. Fixed!

examples/envelope_designer/sketch.js
JunShern: Wow, that essay was eye-opening, and very inspiring! I do feel a definite need to illustrate better how all the different sound components work, but I'm not sure that tacking more onto this example is the best way to do achieve that. 

Let's talk about this more in terms of designing the core introduction-to-p5.js-sound tutorials. :)

Issue #266, opened 2018-03-16
PolySynth docs incorrect


jvntf:
The documentation and inline example indicate that the `note` param should be specified as a MIDI value. This does is not correct-- as it stands, note should be specified as a midi note string or as a frequency. polysynth.play() calls polysynth.noteAttack() which calls [audiovoice._setNote()](https://github.com/processing/p5.js-sound/blob/master/src/audioVoice.js#L27)

JunShern:
Hi @jvntf, I was just having a look at this, I tried changing the example but got confused by some errors, possibly a bug.

To begin with, this doesn't work:

```javascript
var polysynth;
function setup() {
  polysynth = new p5.PolySynth();
  var velocity = 0.7;
  var time = 0;
  polysynth.play("C3",velocity,time,1.5);
  polysynth.play("E3",velocity,time+=0.5,1.0);
  polysynth.play("G3",velocity,time+=0.5,0.5);
  time += 1.0; // Rest
  polysynth.play("G3",velocity,time+=0.5,1.5);
  polysynth.play("E3",velocity,time+=0.5,1.0);
  polysynth.play("C3",velocity,time+=0.5,0.5);
}
```

which gives the error: `Uncaught TypeError: Cannot read property 'triggerAttack' of undefined (sketch: line 11)`.

But when I just change the duration parameter of the first few notes, then it works:

```javascript
var polysynth;
function setup() {
  polysynth = new p5.PolySynth();
  var velocity = 0.7;
  var time = 0;
  polysynth.play("C3",velocity,time,1.0);
  polysynth.play("E3",velocity,time+=0.5,0.5);
  polysynth.play("G3",velocity,time+=0.5,1.0);
  time += 1.0; // Rest
  polysynth.play("G3",velocity,time+=0.5,1.5);
  polysynth.play("E3",velocity,time+=0.5,1.0);
  polysynth.play("C3",velocity,time+=0.5,0.5);
}
```

I'm wondering if this might have something to do with having a limited number of polyphonic voices available, maybe? I could look into it some more but hopefully someone more familiar with it can pick out the problem quickly.

JunShern:
Also to clear this up, both our MonoSynth and PolySynth classes accept the same inputs: either a note string "A#3", or frequency in Hz. If a user wants to work with MIDI numbers, they need to first convert using midiToFreq().

The following example illustrates the three possibilities:

```javascript
var synth;
function setup() {
[...]

PR #280, opened 2018-05-22
monoSynth and polySynth fixes + examples

therewasaguy:
Dug into the synth classes in response to https://github.com/processing/p5.js-sound/issues/266, they have a solid foundation and I'm excited to do more with them!

Apologies for this somewhat large PR, I'll break it down into smaller pieces...

- fix some dispose issues that made examples break if you try to reset them. (I think a better solution for this might be to set a flag on the dispose method so that everything can only be disposed of once)

- update monoSynth and PolySynth inline examples to use Note Strings as discussed in https://github.com/processing/p5.js-sound/issues/266

- fix PolySynth currentVoice counting as discussed in https://github.com/processing/p5.js-sound/issues/266 - the fixes are [here](https://github.com/processing/p5.js-sound/commit/762c76fb1fae9447841eeb84bec687ec68071aa5) and [here](https://github.com/processing/p5.js-sound/commit/1a07a823e588c54de5ae40d34aa5e90aa558c59f) - fixes a bug that sometimes resulted in negative or NaN values

- update p5 lib so that we can use `random(fromArray)` - might wanna move this to a separate PR

- use strings in the synth examples (i.e. "C3") instead of midi numbers which were getting interpreted as really low frequencies

- add inline attack/release examples to synth classes to show how to make them more interactive

I'd like to write some basic tests for the synths, and organize this PR a bit, while taking in any feedback in the meantime.

If you'd like to try the documentation examples in the context of the actual p5js website, try following [this recently updated guide](https://github.com/processing/p5.js-sound/wiki/Documentation#preview-documentation) on the wiki!

### DIFF
--- a/src/monosynth.js
+++ b/src/monosynth.js
@@ -5,8 +5,8 @@ define(function (require) {
   var AudioVoice = require('audioVoice');
 
   /**
-    *  An MonoSynth is used as a single voice for sound synthesis.
-    *  This is a class to be used in conjonction with the PolySynth
+    *  A MonoSynth is used as a single voice for sound synthesis.
+    *  This is a class to be used in conjunction with the PolySynth
     *  class. Custom synthetisers should be built inheriting from
     *  this class.
     *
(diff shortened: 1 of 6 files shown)


therewasaguy:
added some basic tests and they should run in Chrome (had to do https://github.com/processing/p5.js-sound/pull/282 first) - I think this is ready!

JunShern:
Hey @therewasaguy this all looks great, thanks! One thing - did you forget to include the new `p5.js-sound.js` and `p5.js-sound.min.js` files in your PR? I had to build again with `grunt` before I could try out the new fixes.

therewasaguy:
Thanks for the comments, @JunShern !

> did you forget to include the new p5.js-sound.js and p5.js-sound.min.js files in your PR? I had to build again with grunt before I could try out the new fixes.

Yes, I think this makes PR's more readable while they're in review, it's clearer what to focus on when looking at the diff.

> I feel it would be helpful to include the "click to play" text and color-changing visual feedback like we have on some of the other examples. It helps assure visitors that their click is actually doing something, on the off-chance that they cannot hear the sounds for any reason.

Great suggestion. I didn't do this because it would add more code to the documentation example, which should be as focused as possible on the single method or class it is documenting. In fact the code that loads documentation examples will run setup and a default createCanvas method even if there is no setup method on the page, just to allow us to focus on the code we want to document.

At the same time, these examples should be a nice user experience. And I think your suggestion is a good improvement from that perspective.

So yeah, let's go ahead with it! And while we're adding a `setup` method, it'd be nice if _only_ the canvas was clickable. We could do that with an event listener on the canvas p5.Element this [way](https://p5js.org/reference/#/p5.Element/mouseClicked). This would be a good standard for all examples, and we will probably need to redesign all of them to ensure that the user knows they need to click to start the audio context.

Anyway! I just updated the examples for MonoSynth and PolySynth, didn't get to triggerAttack/Release but can do this in a separate PR (or go for it if you're inspired to improve upon these examples!)

> Some of the lower notes (those on Octave 2 in particular) are difficult to hear on laptop speakers, which feels a bit unsatisfying. Maybe shift everything up by an octave?

Hmm, I'll tweak some of the super low ones, I tried to design the examples with laptop speakers in mind and I don't want them to be too piercing.

> In the comments about velocity, let's tell them that the range is from 0 to 1: // velocity (volume, from 0 to 1)

Good call! Updated

PR #288, opened 2018-06-01
Rename envelope

JunShern:
Following my discussion with @therewasaguy , we agreed it would be best to rename the `p5.Env` class to `p5.Envelope` for a consistent naming scheme with other classes in this repository (other classes like Oscillator are written out in full), and to avoid confusion with other common constructs like Environments and so on. 

This PR is not yet ready, I would like to ask for help because I'm not familiar enough with JavaScript to know the best way to do this.

What has been done so far:

- Replaced all occurrences of "Env" in `env.js` with "Envelope"
- Renamed `env.js` to `envelope.js`
- Updated the corresponding "env" entries in `Gruntfile.js` and `src/app.js` to "envelope"

What is left:

- [x] Continue work on p5.Env to replicate p5.Envelope but with added deprecation warning ([started here](https://github.com/JunShern/p5.js-sound/blob/rename-envelope/src/envelope.js#L131))
- [x] Test that the new Envelope class as well as the deprecated Env class both work

In particular, I would be grateful to know what is the best way to "clone" the p5.Envelope class into p5.Env, then I will test and finish up the PR.


src/envelope.js
```diff
@@ -128,9 +128,15 @@ define(function (require) {
     p5sound.soundArray.push(this);
   };
 
+  // THIS DOESN'T WORK YET
+  p5.Env = function(t1, l1, t2, l2, t3, l3) {
+    console.warn('WARNING: p5.Env is now deprecated and may be removed in future versions. Please use the new p5.Envelope instead.');
+    p5.Envelope(t1, l1, t2, l2, t3, l3);
```
therewasaguy: I would think of p5.Env as extending the p5.Envelope similar to the way that these named p5.SqrOsc extends p5.Oscillator https://github.com/processing/p5.js-sound/blob/master/src/oscillator.js#L566 using `call` and Object.create to extend the prototype. In es6 this can be done with `class` and `extends` but we have been using the es5 approach.

You could use the arguments object or the es6 version, using rest parameters + spread operator, when calling

```    p5.Envelope.call(this, ...args);```

**Comment by JunShern** [comment]
Okay! Completed the changes for backwards-compatible p5.Env function which mimics new p5.Envelope, unfortunately we don't have unit tests for Envelope specifically but all other tests (some of which rely on Envelope) and Envelope examples are working correctly. 

I believe this is ready to go! @therewasaguy want to have a look before I merge? Btw adding the `arguments` keyword as the second parameter to `p5.Envelope.call` didn't work, so I just copied out all the arguments again. You can see what I mean in [this commit](https://github.com/processing/p5.js-sound/pull/288/commits/9db7ddd55fad29218726329caea17d02752e999c).

File: src/envelope.js [hunk 2]
```diff
@@ -902,4 +902,12 @@ define(function (require) {
     }
   };
 
+  // Different name for backwards compatibility, replicates p5.Envelope class
+  p5.Env = function(t1, l1, t2, l2, t3, l3) {
+    console.warn('WARNING: p5.Env is now deprecated and may be removed in future versions. ' +
```
therewasaguy: If you'd be up for a quick find-and-replace-all, especially in the `src` and `examples` directories, that'll be helpful to complete the deprecation.

You could do it in one line with a recursive `grep` for `p5.Env(`, piped into `sed`

```
grep -rl "p5.Env(" examples src | xargs sed -i '' 's/p5.Env(/p5.Envelope(/g'
```

...but you'd need to preserve the two `p5.Env`'s in the two lines above

JunShern:
Gotcha! I used my editor's built-in search-and-replace, to replace all Env words with Envelope (it doesn't change Envelope to Envelopeelope). Ran tests again and it's all good. @therewasaguy

PR #292, opened 2018-06-11
synth play method has a default sustain

therewasaguy:
- add a `DEFAULT_SUSTAIN` to the monosynth so that we can call `play('a4')`  without providing any additional parameters

And a few small tweaks:
- move `noteToFreq` to helpers
- fix octave in ANSI note conversion (`value += 12 * (octave -1);`)
- allow for lower case ANSI notes like "a4"
- dispose of polysynth TimelineSignal on noteRelease - otherwise these were not getting garbage collected and a potential memory issue

### DIFF
--- a/src/monosynth.js
+++ b/src/monosynth.js
@@ -3,6 +3,9 @@ define(function (require) {
 
   var p5sound = require('master');
   var AudioVoice = require('audioVoice');
+  var noteToFreq = require('helpers').noteToFreq;
+
+  var DEFAULT_SUSTAIN = 0.15;
 
   /**
     *  A MonoSynth is used as a single voice for sound synthesis.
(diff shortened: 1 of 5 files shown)


src/monosynth.js
```diff
@@ -159,12 +159,9 @@ define(function (require) {
...
   p5.MonoSynth.prototype.triggerAttack = function (note, velocity, secondsFromNow) {
-    var secondsFromNow = secondsFromNow || 0;
-
-    //triggerAttack uses ._setNote to convert a midi string to a frequency if necessary
-    var freq = typeof note === 'string' ? this._setNote(note)
-      : typeof note === 'number' ? note : 440;
-    var vel = velocity || 1;
+    var secondsFromNow = ~~secondsFromNow;
+    var freq = noteToFreq(note);
+    var vel = velocity || 0.1;
```
therewasaguy: set default velocity to 0.1

PR #302, opened 2018-06-25
Examples for Timing and Visualization

JunShern:
Following @therewasaguy and my discussions, I will be adding some example sketches related to timing and visualization for p5 sound generation. The goal is for these examples to eventually live on the [p5 website examples](https://p5js.org/examples/) page, but as a first stop I am adding them to the gh-pages examples.

Current examples include:

- [ ] `array_of_notes` Static song playback from an array of notes.
- [ ] `array_of_notes_soundloop` A more powerful/flexible version of the previous example, which uses the p5.SoundLoop allowing for play/pause/restart as well as interactive tempo control.
- [ ] `visualize_pentatonic` Random pentatonic generation + particle system for note visualization.
- [ ] `virtual_piano` Virtual synth keyboard.
- [ ] `grid_sequencer`: Grid sequencer example using fixed interval where visuals control audio.

This is a work-in-progress. All comments and feedback welcome!


examples/array_of_notes/sketch.js
```diff
@@ -0,0 +1,45 @@
...
+  if (songStarted) {
+    text('song started', width/2, height/2);
+  } else {
+	  text('click to play song', width/2, height/2);
+  }
+}
+
+function touchStarted() {
+  if (!songStarted) { // Only play once
+    for (var i=0; i<song.length; i++) {
```
therewasaguy: I think we could introduce the SoundLoop here. That way the song could be re-started by resetting the index, or paused / stopped.

JunShern: I decided to introduce the SoundLoop in a separate example, because it makes playback somewhat more complicated than necessary depending on what the user needs. So there are now two examples, `array_of_notes` and `array_of_notes_soundloop` which illustrate two possible ways to play music from an array.

JunShern:
Hi all! I've made some updates to all the examples, cleaning up the code and adding descriptions to the top of each sketch. 

I think these are good to go now, if there are no objections I think we can go ahead and merge these into the main branch, and then I'll start the request to add them to the main p5.js website. :smile: 

@therewasaguy

examples/grid_sequencer/sketch.js
```diff
@@ -0,0 +1,162 @@
...
+  this.width = cellWidth - 2 * this.padding;
+  this.height = cellHeight - 2 * this.padding;
+  this.defaultColor = [190, 240, 255];
+  // Mouse hover
+  this.hovered = false;
+  this.hoverColor = [230, 255, 255];
+  // Enabled when clicked
+  this.enabled = false;
+  var varyingColorVal = 22 * (this.pitch % pitches.length);
+  this.enabledColor = [20 + varyingColorVal, 255 - varyingColorVal, 255];
```
therewasaguy: One of the `enabledColor` is hard for me to distinguish from the `defaultColor`. It should be able to work in greyscale—there are chrome extensions like [this](https://www.funkify.org) to test this kind of thing.

`colorMode(HSB)` might make this color scheme easier, where "disabled" always has a different brightness than "enabled"

Issue #301, opened 2018-06-24
Inappropriate MonoSynth velocity range causes clipping/distortion


JunShern:
I've noticed that when using MonoSynth on default settings, any velocity above 0.4 seems to introduce clipping. This is heard as a distorted voice, instead of the smooth sine tone we expect. Visually, an FFT waveform visualizer confirms that there is clipping when velocity goes about 0.4:

https://alpha.editor.p5js.org/JunShern/sketches/r1UjCwhWQ

For a single voice, the distortion is not so obvious but it causes problems down the line in PolySynth as well, which are less bearable:

https://alpha.editor.p5js.org/JunShern/sketches/BJc9TDnWQ

And for polyphony (about 4-ish notes) it seems like keeping the velocity below 0.1 seems to reduce distortion. However, maybe we can fix the range for MonoSynth first, then handle suppression for multiple notes in PolySynth in a separate issue.

Does anyone know if this range issue is a bug in the math/code for MonoSynth, or is it just something we should adjust empirically?

@therewasaguy @jvntf

PR #305, opened 2018-06-26
patch monosynth velocity

therewasaguy:
This should fix https://github.com/processing/p5.js-sound/issues/301 
- Monosynth envelope ramps `this.output.gain` — the audio param — instead of `this.output`
- [Update master limiter knee and threshold](https://github.com/processing/p5.js-sound/commit/9607a544e94246649f3a724425121762786dcae6) so that it is better suited to prevent clipping. Specifying a sharp `knee` value makes a big difference, as does lowering the ratio slightly. We don't want to change the quality of sound unless it's nearly clipping.
- remove some unused monosynth properties (_isOn and filter)

### DIFF
--- a/src/master.js
+++ b/src/master.js
@@ -12,8 +12,9 @@ define(function () {
 
     //put a hard limiter on the output
     this.limiter = audiocontext.createDynamicsCompressor();
-    this.limiter.threshold.value = 0;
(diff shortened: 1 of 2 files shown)


JunShern:
Awesome, this looks and sounds great! Should be good to merge. :+1:

PR #308, opened 2018-07-26
Fractal music

JunShern:
New example demonstrating the use of fractal systems (in particular using a Lindenmayer system to generate fractals) to generate musical content.


examples/fractal_music/sketch.js
```diff
@@ -1,94 +1,109 @@
...
 var rules = {
-  "A":["A","B"],
-  "B":["E","C"],
-  "C":["F","G"],
-  "D":["D","F"],
-  "E":["G"],
-  "F":["B","A"],
-  "G":["C"]
+  "+":["+","-"],
+  "-":["+"]
```
therewasaguy: could we use numbers to represent these intervals?

semitone intervals are not so pleasant, have you tried whole tone, fourth, fifth?

this pattern will wind up increasing, how to address this?

JunShern:
Hey @therewasaguy! I've been playing around with this some more and I feel that maybe the using A,B,C symbols instead of +/- is a better way to go, considering the problem we discussed about the melody climbing or falling to extreme pitches as the fractal advances.

I'm also using randomness to choose from a set of common note durations, just to give it more character.

I think it sounds decent now, and am happy to keep this example unless/until someone figures out how to make it sound better!

examples/fractal_music/sketch.js
```diff
@@ -0,0 +1,141 @@
...
+    sequences.shift(); // Removes first element from array
+  }
+}
+
+function highlightNote(seqInd, noteInd, highlightColor) {
+  if (noteInd < 0) return; // Skip if we haven't started
+  // X position of character
+  var seqWidth = textWidth(sequences[seqInd].join(" "));
+  var xOffset = width/2 - seqWidth/2;
+  var x = xOffset + seqWidth * noteInd / sequences[seqInd].length;
```
therewasaguy: hmm, the x position is a bit off when notes aren't centered. Maybe the math would be easier with `rectMode(CENTER)` ?

PR #312, opened 2018-08-05
Added saveSoundToBlob() to soundRecorder.js

wahengchang:
as my need is recording sound by microphone, then appending it to DOM for preview before it is downloaded. a temporary blob url is the choice.

### DIFF
--- a/src/soundRecorder.js
+++ b/src/soundRecorder.js
@@ -245,6 +245,18 @@ define(function (require) {
   };
 
   /**
+   *  Export a usable url for blob object.
+   *
+   *  @method saveSoundToBlob
+   *  @param  {p5.SoundFile} soundFile p5.SoundFile that you wish to export
+   */
+  p5.prototype.saveSoundToBlob = function(soundFile) {
+    const dataView = convertToWav(soundFile)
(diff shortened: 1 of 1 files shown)


therewasaguy:
thank you @wahengchang ! This will close https://github.com/processing/p5.js-sound/issues/257

therewasaguy:
as part of a related issue (https://github.com/processing/p5.js-sound/issues/314) we may want to move these away from the p5.prototype to the SoundFile

wahengchang:
@therewasaguy , when can I use the new version which included this Pull-Request ?
I need it desperately

therewasaguy:
@wahengchang you can use it now by compiling p5 sound from source, and it will be included in the next release of p5.js, tho I'd like to move this (and saveSound) to p5.SoundFile / p5.AudioIn, before we make it part of the official API. Let me know if you have any thoughts on that, I'm about to take a look!

therewasaguy:
^ https://github.com/processing/p5.js-sound/pull/315 would move `saveBlob` to SoundFile. I think that's the best home for this method?

wahengchang:
👍

PR #315, opened 2018-08-11
add `save()` and `saveBlob()` methods to SoundFile. Move .wav conversion to helpers.js

therewasaguy:
- move .wav conversion from p5.SoundRecorder to helpers module
- add  `save` and `saveBlob` method to SoundFile
- document the global `saveSound` method (it was listed under p5.SoundRecorder but is actually part of p5.prototype)
- p5.SoundFile's `saveBlob` method can probably replace the global `saveSoundAsBlob` added in https://github.com/processing/p5.js-sound/pull/312 —@wahengchang does this sound like a good approach to you?

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -70,7 +70,7 @@ define(function (require) {
   };
 
   // This method converts ANSI notes specified as a string "C4", "Eb3" to a frequency
-  noteToFreq = function(note) {
+  var noteToFreq = function(note) {
     if (typeof note !== 'string') {
       return note;
     }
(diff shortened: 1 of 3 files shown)


therewasaguy:
Thank you @wahengchang! I like `getBlob` that's a better name than `saveBlob`. I wonder if it is really necessary to have `getBlobUrl` if the audio is already in a SoundFile, maybe we can let users do that with `URL.createObjectURL` on their own if/as needed?

wahengchang:
agree, `getBlob()` is good enough !

PR #320, opened 2018-09-10
prevent crashes during completion of callbacks on disposed SoundFiles

Spongman:
- adds some checks to SoundFile callbacks to prevent crashes during async callbacks completing on disposed objects
- moves `loadSound` call to `preload()` in doc sample

### DIFF
--- a/src/soundfile.js
+++ b/src/soundfile.js
@@ -221,9 +221,11 @@ define(function (require) {
       request.onload = function() {
         if (request.status === 200) {
           // on sucess loading file:
+          if (!self.panner) return;
           ac.decodeAudioData(request.response,
             // success decoding buffer:
             function(buff) {
+              if (!self.panner) return;
               self.buffer = buff;
               self.panner.inputChannels(buff.numberOfChannels);
               if (callback) {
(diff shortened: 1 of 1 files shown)


src/soundfile.js
```diff
@@ -221,9 +221,11 @@ define(function (require) {
       request.onload = function() {
         if (request.status === 200) {
           // on sucess loading file:
+          if (!self.panner) return;
```
therewasaguy: This does the trick, but I wonder if there's a better way to check if the soundfile has been disposed. Or, better yet, to cancel any incomplete callbacks when the soundfile is disposed.

src/soundfile.js
```diff
@@ -1660,9 +1660,10 @@ define(function (require) {
    *  <div><code>
    *  var inp, button, mySound;
    *  var fileName = 'cool';
-   *  function setup() {
+   *  function preload() {
    *    mySound = loadSound('assets/doorbell.mp3');
-   *
+   *  }
+   *  function setup() {
```
therewasaguy: 👍 sorry for the trouble here!

PR #323, opened 2018-10-02
removed line from p5.Filter.prototype.freq which sets the frequency b…

clarkenheim:
…efore the exponential ramp, which was essentially ignoring the 'time' argument passed to the method


therewasaguy:
thanks @clarkenheim !

Issue #341, opened 2019-02-27
SoundFile always saves as 44.1kHz instead of the sampling rate of the AudioContext


codygratner:
Lines 824-825 from p5.sound.js 0.3.10 (from the p5.js 0.7.3 release) has the sample rate hard coded at 44100 when saving a p5.SoundFile.

```
view.setUint32(24, 44100, true);
view.setUint32(28, 44100 * 4, true);
```

My audio interface is set to 48kHz and doesn't want to change. When using:

```
this.recorder = new p5.SoundRecorder();
this.soundFile = new p5.SoundFile();
this.recorder.record(this.soundFile);
this.recorder.stop();
save(this.soundFile, 'mySound.wav');
```

I end up with a wave file that is set to 44100, but correctly captures the 48000 sampling rate audio. I just open it in Wavosaur and change the sample rate to 48000 with no issues. But, those two lines can be changed to the following to fix that:

```
view.setUint32(24, p5sound.audiocontext.sampleRate, true);
view.setUint32(28, p5sound.audiocontext.sampleRate * 4, true);
```

I have no idea if that would break anything else? I saw issue #107 that didn't look related to this?

therewasaguy:
Thanks and good call, @codygratner! Feel free to make a pull request with the fix

PR #342, opened 2019-02-27
Used sampleRate from AudioContext for saving rather than hardcoded 44100

codygratner:
The ```convertToWav(audioBuffer)``` function hardcoded 44100 for the sample rate, which saved the buffer to a file correctly, but set the sample rate wrong if the buffer's sample rate was different (e.g., 48000). One could just alter the file after, but this saves the buffer to a wave file with the sampleRate from the AudioContext.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -259,8 +259,8 @@ define(function (require) {
     view.setUint16(20, 1, true);
     // stereo (2 channels)
     view.setUint16(22, 2, true);
-    view.setUint32(24, 44100, true);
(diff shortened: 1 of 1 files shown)


therewasaguy:
Fixes https://github.com/processing/p5.js-sound/issues/341

thanks @codygratner!

codygratner:
Thanks @therewasaguy ! This was my first fork and first pull request and first contribution to an open source project ... this was minor, but I feel accomplished ... AND, it fixed my issue with wave files at the wrong sample rate!

PR #347, opened 2019-03-03
Changing all var to let in src folder

### LINKED ISSUE #345: Migrating src code to ES6
I am working on it :)

iFlameing:
Fixes #345

### DIFF
--- a/src/amplitude.js
+++ b/src/amplitude.js
@@ -1,7 +1,7 @@
 'use strict';
 
 define(function (require) {
-  var p5sound = require('master');
+  let p5sound = require('master');
 
   /**
    *  Amplitude measures volume between 0.0 and 1.0.
@@ -15,7 +15,7 @@ define(function (require) {
    *                             amplitude readings (defaults to 0)
    *  @example
    *  <div><code>
-   *  var sound, amplitude, cnv;
+   *  let sound, amplitude, cnv;
    *
    *  function preload(){
    *    sound = loadSound('assets/beat.mp3');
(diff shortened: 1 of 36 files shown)


iFlameing:
@therewasaguy can you review this :)

therewasaguy:
Hi @iFlameing, thanks for starting on this ambitious issue #345! I think it'll take more than a find + replace to bring us to ES6 (and beyond), but it's a nice start.

I'm wondering if we should consider using `const` as part of this. I don't know why p5.js isn't using `const`. Do you know? There are a lot of places where `const` would be a better fit because the reference is not going to change.

iFlameing:
@therewasaguy I am waiting for the reply nice to hear from you thanks! I will try to convert all not used let into const.. Can you also review the other pr?

therewasaguy:
@iFlameing did you try compiling this before submitting the PR?

It doesn't compile for me, I'm getting this error on the require.js minification step:

```
Running "requirejs:unmin" (requirejs) task
Running "requirejs:min" (requirejs) task
Error: Cannot uglify2 file:
>> p5.js-sound/lib/p5.sound.min.js. Skipping it. Error
>> is:
>> Unexpected token: name (node) (line: 98, col: 20, pos: 4897)
>> 
```

I'm guessing it has to do with `let`'s impact on variable scope, a quick search brings up this issue
https://github.com/requirejs/requirejs/issues/693

The non-minified build succeeded, though!

iFlameing:
@therewasaguy sorry I missed it. I am also getting the same error. I will try to fix it :)

iFlameing:
@therewasaguy I saw modified file and found that due to changing var to let some variable becomes undefined due to scope issue of let keyword. what you suggest me to do in that case. should I only convert those var in let which is correct for both or I have to modify each file and each variable which is not defined after the changing the var to let? what do you think?

iFlameing:
@therewasaguy I found this interesting [issue](https://github.com/webpack/webpack/issues/2972) which tells that we can't use let keyword in our code because uglify only understand es5 so we have to use babel for convert es6 to es5.

iFlameing:
@therewasaguy Fixes the minimization issue https://github.com/processing/p5.js-sound/pull/349

therewasaguy:
in progress here https://github.com/processing/p5.js-sound/pull/366

PR #349, opened 2019-03-07
Fixing minifying problem from es6 to es5

### LINKED ISSUE #348: upgrade build tools so that we can use ES6+ conventions in source code

iFlameing:
Fixes #348  (upgrade build tools so that we can use ES6+ conventions in source code)

@therewasaguy can you review this?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -142,8 +142,14 @@ module.exports = function(grunt) {
               });
            }
           },
-          optimize: 'uglify2',
-          out: 'lib/p5.sound.min.js',
+          optimize: 'none',
+          out: function(text, sourceMapText) {
+
+            var UglifyJS = require('uglify-es'),
+                uglified = UglifyJS.minify(text);
+
+            grunt.file.write('lib/p5.sound.min.js', uglified.code);
+          },
           paths: '<%= requirejs.unmin.options.paths %>',
           useStrict: true,
           wrap: {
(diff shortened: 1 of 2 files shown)


therewasaguy:
Nice work, @iFlameing !

We're not able to use some ES6 conventions—for example, arrow functions (`const myArrow = () => {}`) will result in the following error in grunt's requirejs:unmin step
```
{ Error: Error: Line 10: Unexpected token )
    at throwError (/Users/jasonsigal/src/p5/p5.js-sound/node_modules/esprima/esprima.js:1846:21)
```

It seems this is due to `amdclean` (a library that has not been updated since 2016) and its dependency, `esprima`. If we comment this out, it builds (but without the formatting that we want)

```
               'escodegen': {
                  'comment': false,
                  'skipDirOptimize':true,
                  'format': {
                    'indent': {
                      'style': '  ',
                      'adjustMultiLineComment': true
                    }
                  }
                }
```

Before we get too deep into this, it might be worth considering, as part of this effort, some alternatives to require.js / AMD which sort of lost the JavaScript module system showdown (and that's probably why the amdclean library hasn't been updated in 3 years). One option is Browserify and Babelify. This is what p5.js uses now, and I've started investigating in [this branch](https://github.com/processing/p5.js-sound/compare/master...therewasaguy:babelify?expand=1) but it's kind of a mess. Another option is webpack but I investigated that once and it didn't play nice with [the way we generate documentation](https://github.com/processing/p5.js-sound/wiki/Documentation) from inline comments in the unminified source code.

therewasaguy:
in progress here https://github.com/processing/p5.js-sound/pull/366

Issue #88, opened 2015-12-13
p5.SoundFile dispose() method doesn't free up resources


bplmp:
When loading many p5.SoundFiles (more than 150 in my case) the browser becomes slow. 

I've tried using the dispose() method or assigning the sound objects to null, but nothing seems to free up the memory after the sounds have been loaded. Any way to do this?

p5 forum topic for reference: https://forum.processing.org/two/discussion/comment/57306

brig002:
same problem here. but in my case, just playing same sound file ~20 times in quick succession freezes browser (my sound file is less than a second). can happen even after sounds have stopped playing. reference post: https://forum.processing.org/two/discussion/14223/running-sound-play-more-than-20-times-slows-browser-to-a-crawl#latest

therewasaguy:
Hey @bplmp and @brig002, thanks for reporting. I've been meaning to dig into this and I'm sorry I have not chimed in.

I'm pretty sure that the issue has to do with two things:

1. a potential memory leak in the SoundFile

2. The way that SoundFile keeps track of currentTime. This is not built in to the Web Audio API's BufferSourceNode, so it uses an additional ScriptProcessorNode to track the position in the buffer. That node is bad news and I think we have to get rid of it. I've noticed that the problems are particularly bad in firefox.

   Unfortunately, I could not find a better way to accurately track current time of a playback buffer in the web audio api in cases where the user changes the playback rate. For more, you can see [this thread](http://stackoverflow.com/questions/21816484/how-i-can-get-current-time-when-use-playbackrate-value-in-web-audio-api#comment33050207_21816484).

   I think the library should either drop support for SoundFile.currentTime, or make a note that this method will not work accurately if the playback rate is changed. If users want accurate playback, the documentation should encourage use of the p5.dom library's createAudio method to create an HTML5 Audio Element. That element is designed to track playback position, while Web Audio API is designed for sample-accurate scheduling. As part of this, it would be good to make a tutorial that outlines when to use which type of audio player...

therewasaguy:
Looking into the memory issues, [this](http://stackoverflow.com/a/25509604/2994108) is an informative thread. Excerpt:

"memory is large because the Web Audio API decodes your small MP3 into 32-bit LPCM – which will give you something on the order of 10MB per minute per channel"

@bplmp when you load 150 sound files, are any of them referencing the same file? Or are they all unique audio clips? How long are the clips?

brig002:
[...]

PR #350, opened 2019-03-13
Fix memory leak in soundfile playback (#88, #100)

therewasaguy:
a long overdue fix for #88 and #100


therewasaguy:
🚢

joemckay5:
Just had a similar "crackling" sound issue in chrome and found rolling back the sound to 0.9.0 fixed the problem.

PR #364, opened 2019-05-28
Add tests for p5.SoundRecorder and lint existing tests

oshoham:
- Add new test file for p5.SoundRecorder
  - Test recording from a microphone
  - Test recording from a sound file
  - Test recording from the master output
  - Test saving a sound file as a .wav
- Add `package-lock.json` (the main p5 repo has one so I wasn't sure whether to exclude it or not)
- Move `.estlintrc` to the top level of the repo so that it can be used when editing files in both `src` and `test`
- Fix linting errors in existing tests
- Remove broken/outdated test from p5.AudioVoice.js test file


test/tests/p5.SoundRecorder.js
```diff
@@ -0,0 +1,142 @@
...
+
+      // need to enable master volume to test recording from the microphone
+      p5.prototype.masterVolume(1);
+
+      var mic = new p5.AudioIn();
+      mic.start(function() {
+        var outputSoundFile = new p5.SoundFile();
+        recorder.record(outputSoundFile);
+
+        setTimeout(function() {
```
therewasaguy: we discussed changing setTimeout to a callback to the record method

test/tests/p5.SoundRecorder.js
```diff
@@ -0,0 +1,142 @@
...
+      recorder.dispose();
+    });
+
+    it('can record input from a microphone', function(done) {
+      var recordingDuration = 100;
+
+      // need to enable master volume to test recording from the microphone
+      p5.prototype.masterVolume(1);
+
+      var mic = new p5.AudioIn();
```
therewasaguy: we discussed adding an `await` for mic access if we return the promise here

https://github.com/processing/p5.js-sound/blob/master/src/audioin.js#L133

test/tests/p5.SoundRecorder.js
```diff
@@ -0,0 +1,142 @@
...
+      mic.start(function() {
+        var outputSoundFile = new p5.SoundFile();
+        recorder.record(outputSoundFile);
+
+        setTimeout(function() {
+          recorder.stop();
+          expect(outputSoundFile.duration()).to.be.closeTo(recordingDuration / 1000, 0.1);
+
+          var outputChannel = outputSoundFile.buffer.getChannelData(0);
+          expect(outputChannel[0]).to.not.eq(0);
```
therewasaguy: potentially flaky but I can't think of a better way

test/tests/p5.SoundRecorder.js
```diff
@@ -0,0 +1,142 @@
...
+          done();
+        }, recordingDuration);
+      });
+    });
+
+    it('can record input from a sound file', function(done) {
+      // TODO: why does a sampleIndex of 0 break this test?
+      var sampleIndex = 1000;
+      var recordingDuration = 100;
+      var inputChannel = inputSoundFile.buffer.getChannelData(0);
```
therewasaguy: we discussed making the buffer rather than loading a .wav file

test/tests/p5.SoundRecorder.js
```diff
@@ -0,0 +1,142 @@
...
+          mic.dispose();
+          p5.prototype.masterVolume(0);
+          done();
+        }, recordingDuration);
+      });
+    });
+
+    it('can record input from a sound file', function(done) {
+      // TODO: why does a sampleIndex of 0 break this test?
+      var sampleIndex = 1000;
```
therewasaguy: this is a good question! Might have to do with scriptProcessorNode's 1024 bufferSize but we're not sure yet

test/tests/p5.SoundRecorder.js
```diff
@@ -0,0 +1,142 @@
...
+      inputSoundFile.play();
+      recorder.setInput();
+      recorder.record(outputSoundFile);
+
+      setTimeout(function() {
+        recorder.stop();
+        expect(outputSoundFile.duration()).to.be.closeTo(recordingDuration / 1000, 0.1);
+
+        var outputChannel = outputSoundFile.buffer.getChannelData(0);
+        expect(outputChannel[0]).to.not.eq(0);
```
therewasaguy: maybe we can be more precise here if it's a generated buffer

test/tests/p5.SoundRecorder.js
```diff
@@ -0,0 +1,142 @@
...
+
+  describe('p5.SoundRecorder', function() {
+    var recorder;
+    var inputSoundFile;
+    var originalWriteFile;
+    var writeFileCalled;
+
+    before(function(done) {
+      // clean up audio nodes from other test suites
+      p5.soundOut.soundArray.length = 0;
```
therewasaguy: nice, you found the soundArray!

Setting length to 0 isn't something I've seen before. Will this free up the resources for garbage collection?

An approach we've taken is to explicitly call `dispose` on the nodes in this array. But if we are properly disposing at the end of each test, this may not be necessary?

test/tests/p5.SoundRecorder.js
```diff
@@ -0,0 +1,142 @@
...
+
+      // mock p5.prototype.writeFile
+      originalWriteFile = p5.prototype.writeFile;
+      p5.prototype.writeFile = function() {
+        writeFileCalled = true;
+      };
```
therewasaguy: let's check if p5 is using sinon

oshoham: it looks like p5 does use sinon a bit, but only to stub one specific function:

https://github.com/processing/p5.js/blob/2d4df80e91879c71e34c0fb2553554315d0d4156/test/unit/io/loadImage.js#L6

oshoham: I'm happy to use sinon here if you're ok with it - it's probably better long-term to use a real mocking library

test/tests/p5.SoundRecorder.js
oshoham: after giving it some thought, I'd rather not modify `p5.AudioIn` as part of this PR. instead, I'll add a request for mic access to the `before` hook for this test suite, and throw an error if mic access is denied

test/tests/p5.SoundRecorder.js
oshoham: I think it's better to assume that other tests are properly disposing of their resources, so I've taken this line out entirely.

test/tests/p5.SoundRecorder.js
oshoham: I never did figure out why this was happening, but switching to a generated input buffer and recording for the shortest possible duration (the recorder's bufferSize / sampleRate) seems to have removed the need for this workaround.

test/tests/p5.SoundRecorder.js
oshoham: done!

test/tests/p5.SoundRecorder.js
oshoham: still can't think of a better assertion for this, but I'm happy to change it in the future if we come up with something?

test/tests/p5.SoundRecorder.js
oshoham: done!

test/tests/p5.SoundRecorder.js
oshoham: Even with a generated buffer, I'm still having issues testing against precise values here. It seems like playing the input SoundFile through the master soundOut node produces a different output buffer, even with the master volume at 1. I don't really understand why...

therewasaguy: Ah, we have a pretty aggressive dynamicCompressorNode inside of the Master that is probably clamping the values. Perhaps a value of 0.5 would pass thru just fine?

therewasaguy: Ah, we have a pretty aggressive dynamicCompressorNode inside of the Master that is probably clamping the values. Perhaps a value of 0.5 would pass thru just fine?

test/tests/p5.SoundRecorder.js
therewasaguy: 👍

test/tests/p5.SoundRecorder.js
therewasaguy: 👍

test/tests/p5.SoundRecorder.js
oshoham: Ah, that's probably it.

oshoham:
@therewasaguy Yeah I agree that it would probably be better to import sinon as an npm dependency, but I think that that would require some restructuring of how the tests are run (or maybe I just don't understand Grunt well enough?) Maybe in a separate PR?

PR #373, opened 2019-07-10
Replace ScriptProcessorNode with AudioWorkletNode in p5.SoundFile and p5.Amplitude

oshoham:
* Replace ScriptProcessorNode functionality with equivalent AudioWorklet processors in p5.SoundFile and p5.Amplitude
* Ensure that a global p5 instance is created and initialized before the test suite runs, so that all AudioWorklet modules are loaded before the tests run.


src/amplitude.js
```diff
@@ -50,38 +51,42 @@ define(function (require) {
 
     // set audio context
     this.audiocontext = p5sound.audiocontext;
-    this.processor = this.audiocontext.createScriptProcessor(this.bufferSize, 2, 1);
+    this._workletNode = new AudioWorkletNode(this.audiocontext, processorNames.amplitudeProcessor, {
+      outputChannelCount: [1],
+      parameterData: { smoothing: smoothing || 0 },
```
therewasaguy: as a potential improvement, we could check the range here, as in the `smooth` method. But we could also just let the parameterDescriptors handle enforcing (and choosing whether to notify the user) about the range.

src/audioWorklet/amplitudeProcessor.js
```diff
@@ -0,0 +1,93 @@
...
+
+class AmplitudeProcessor extends AudioWorkletProcessor {
+  static get parameterDescriptors() {
+    return [
+      {
+        name: 'smoothing',
+        defaultValue: 0,
+        minValue: 0,
+        maxValue: 1,
+        automationRate: 'k-rate'
```
therewasaguy: inspecting this in Chrome 75.0.3770.100, somehow it has become `'a-rate'`. Confusing, but looks like you're following the spec https://webaudio.github.io/web-audio-api/#ref-for-enumdef-automationrate%E2%91%A1

oshoham: yeah, I see the bug, it looks like the polyfill is assuming that all worklet parameters are `a-rate`: https://github.com/GoogleChromeLabs/audioworklet-polyfill/blob/d6f1a14dacc597876b1a5f9cd8719729a131b585/src/index.js#L96

I think your suggested approach makes sense - another possible solution is to treat the smoothing parameter as `a-rate` (i.e. a Float32 buffer of values) and update the processor code accordingly. This seems more in line with the intended use case of AudioParams, although probably less efficient for non-AudioWorklet browsers. What's your opinion?

PR #377, opened 2019-08-08
remove webpack comments from build (for yuidoc)

therewasaguy:
p5 generates p5.sound documentation by looking at the p5.sound.js file in the addons directory [here](https://github.com/processing/p5.js/blob/3da1b0286a7b99d4342e209ee1ba3f3286d5dff9/Gruntfile.js#L235).

Unfortunately, our new webpack build adds comments that confuse the yuidoc parser.

@oshoham and I spent some time looking at possible solutions. We tried to remove extraneous comments from the webpack build, but this caused more problems.

The solution we came to for now is to separate out the comments using the "extractComments" options, which is designed to extract a license file. We'll include this file with updates to the p5 repo. Then p5 will pick up the comments from that file when it generates the reference pages.

We'd need to either remove comments from the un-minified file, or change [this](https://github.com/processing/p5.js/blob/3da1b0286a7b99d4342e209ee1ba3f3286d5dff9/Gruntfile.js#L235) to point to a documentation directory that only has this documentation file.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -4,7 +4,22 @@ module.exports = function(grunt) {
 
   grunt.initConfig({
     pkg: grunt.file.readJSON('package.json'),
-
+    decomment: {
+      any: {
+        // remove comments added by webpack from the build
+        files: {
+            "./lib/p5.sound.js": "./lib/p5.sound.js",
+        },
+        options: {
+          ignore: [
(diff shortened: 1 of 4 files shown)


therewasaguy:
Update:

We can just remove the comments that webpack adds, but keep the JSDoc comments, with help from https://www.npmjs.com/package/decomment and some regex

PR #378, opened 2019-08-14
Remove bad console call

arist0tl3:
Correct console function call


therewasaguy:
ahh, thank you @arist0tl3

arist0tl3:
@therewasaguy 

Hey, thanks for the quick merge. I'm wondering what the process is to get something like https://cdnjs.com/libraries/p5.js/ updated with this fix, as we have a few projects pointing at this and it would be nice to not have to use local code fixes for them =)

Thanks!

PR #380, opened 2019-08-15
Bugfixes for p5.Amplitude and p5.Soundfile for browsers without AudioWorklet support

oshoham:
As pointed out by @therewasaguy, there is a bug in the AudioWorklet processor for p5.Amplitude in Firefox and Safari, in which the `smoothing` AudioParam is being [passed to the processor as a Float32 array by the AudioWorklet polyfill](https://github.com/GoogleChromeLabs/audioworklet-polyfill/blob/d6f1a14dacc597876b1a5f9cd8719729a131b585/src/index.js#L96), rather than as a single value. This PR fixes that issue by setting the `smoothing` parameter via the worklet's message port instead of using an AudioParam.

This PR also fixes an issue with pausing p5.SoundFiles, in which the SoundFile AudioWorklet processor was incorrectly setting the SoundFile's playback position to 0 on pause.


src/audioWorklet/soundFileProcessor.js
```diff
@@ -2,12 +2,29 @@
...
+  }
+
   process(inputs) {
     const input = inputs[0];
     const inputChannel = input[0];
     const position = inputChannel[inputChannel.length - 1] || 0;
 
-    this.port.postMessage({ name: 'position', position: position });
+    if (!this.paused) {
+      this.port.postMessage({ name: 'position', position: position });
```
therewasaguy: it seems we can't rely on `this.paused` having updated in Safari. Maybe this logic needs to be in the SoundFile?

src/soundfile.js
```diff
@@ -539,6 +539,8 @@ define(function (require) {
     var pTime = time + now;
 
     if (this.isPlaying() && this.buffer && this.bufferSourceNode) {
+      this._workletNode.port.postMessage({ name: 'pause' });
```
therewasaguy: in Safari, this message doesn't always update the worker's `paused` attribute before it can post a `position` message of 0. As a result, `SoundFile._lastPos` gets set to 0 and so `SoundFile.currentTime()` returns 0 while the soundfile is paused. It's inconsistent, happens maybe half the time, and only in Safari.

It resumes playback from the proper position because `pauseTime` gets set below. Maybe we also/instead need to set a paused attribute on the SoundFile rather than the Processor?

src/soundfile.js
```diff
@@ -539,6 +539,8 @@ define(function (require) {
...
+
       this.pauseTime = this.currentTime();
       this.bufferSourceNode.stop(pTime);
       this._counterNode.stop(pTime);
```
therewasaguy: one issue we should be aware of is that a pause can be scheduled in the future...perhaps a reason to make `pause` an a-rate param so that we could setValueAtTime?

src/audioWorklet/soundFileProcessor.js
oshoham: sure, it's easy enough to move this logic to the SoundFile class

src/soundfile.js
oshoham: if we move the pause logic out of the worklet processor and into the SoundFile, is that still an issue?

therewasaguy: Yes, unfortunately that wouldn’t solve the issue. It was an issue before this pr. Just thinking whether keeping pause logic in the processor could help fix it.

src/soundfile.js
oshoham: The solution I arrived at was to remove the pause logic from the processor and add a guard statement in the worklet node's port callback that returns early if the SoundFile is paused or the reported playback position is 0, which handles the race condition that you encountered in Safari.

Because the processor always looks for the last position value in the current frame of the counter buffer, the playback position will only ever be 0 if the SoundFile is paused (i.e., on the 1st frame, the playback position will be 128/256/whatever the buffer size of the worklet processor is).

oshoham: https://github.com/processing/p5.js-sound/pull/380/commits/ce48b1b4c12241fc2ce3445331fe9a06c502f872#diff-635fd3ddb6f066be99025f1994db3e30R1251

src/soundfile.js
oshoham: I tried making `pause` an a-rate AudioParam that would get converted to a boolean in the worklet processor, but I had trouble getting that approach to work properly, even in Chrome. For some reason, `setValueAtTime` wasn't changing the value of the AudioParam.

One way to handle this problem with the approach that I settled on (see the comment thread above) could be to modify this [guard statement](https://github.com/processing/p5.js-sound/commit/ce48b1b4c12241fc2ce3445331fe9a06c502f872#diff-635fd3ddb6f066be99025f1994db3e30R1251) to only check `event.position.data === 0` rather than `this._paused || event.position.data === 0`.

`event.data.position` should only equal 0 when the counter buffer has actually stopped playing, which would be compatible with scheduling a pause in the future. I'm not 100% sure this would work, but it's my best guess...

src/soundfile.js
therewasaguy: > the playback position will only ever be 0 if the SoundFile is paused

I'm not sure if this is the case in Firefox, where if you pause and un-pause rapidly, the soundfile will stutter at the same position (including position 0), rather than advancing one buffer frame at a time. That's a tiny bug but probably not a blocker on this PR.

src/soundfile.js
therewasaguy: nice solution! At some point, I'd be curious to experiment more with `setValueAtTime` on worklet AudioParams, but this works for now.

Either way, it seems like we only need to check `event.position.data === 0` in the guard statement?

examples/pause_soundfile/sketch.js
```diff
@@ -14,16 +15,35 @@ function setup() {
...
+  if (getAudioContext().state === 'running') {
+    soundFile.loop();
+  }
+
   createP('Press any key to pause. Resume when the key is released')
 }
 
+function touchStarted() {
+  if (getAudioContext().state !== 'running') {
+    getAudioContext().resume();
```
therewasaguy: We have a `userStartAudio` method that can be useful for this, too https://p5js.org/reference/#/p5.sound/userStartAudio

oshoham: I've updated the sketch to use `userStartAudio` instead.

src/soundfile.js
oshoham: I believe so, yes. I changed the guard statement to only check `event.position.data === 0` and it seems to work on Chrome, Firefox, and Safari.

PR #381, opened 2019-09-04
only add audioworklet modules once

therewasaguy:
This Promise `return ac.audioWorklet.addModule(objectURL)` was throwing an error if the worklet with that name had already been registered on the page.

We run into this issue when p5 is in "instance mode" (i.e. on the documentation examples at p5js.org) because the "init" callback can run multiple times for each instance.

cc @oshoham


oshoham:
@therewasaguy Thanks for catching this! Sorry for taking a while to review this PR, I missed the notification.

PR #395, opened 2019-12-11
Fix wrong `splice` method call in `_clearOnEnd` function

SegaraRai:
Fix wrong `splice` method call in `_clearOnEnd` function.


SegaraRai:
Sorry I found a bug in my code.
I'll fix it and reopen PR.

SegaraRai:
Opened #396

PR #396, opened 2019-12-11
Fix deletion of array elements in iteration

SegaraRai:
Fix deletion of array elements in iteration.

- Add the second argument `1` for `splice`
- Iterate in reverse order since the index changes by `splice`

[see this logic works](https://jsfiddle.net/osfdt0z6/)


therewasaguy:
nice, thank you @SegaraRai!!

PR #403, opened 2020-01-03
update inline examples (#388)

therewasaguy:
- Update inline examples to start audio context on a user gesture, either by playing a sound (`soundFile.play()` / `oscillator.start()`) or by explicitly calling `userStartAudio()` https://github.com/processing/p5.js-sound/issues/388
- Use `let` in examples, instead of `var`, but not `const`, as discussed in this issue where it was decided to use `let` exclusively in examples: https://github.com/processing/p5.js/issues/3877
- Update styles for consistency with other p5 inline examples

So far, some examples use `soundFile.play()` or `oscillator.start()` rather than a potentially redundant call to `userStartAudio()`. It _might_ be worth that redundant call, because the difference between methods that call "start" on a WebAudio node (thus enabling audio on a user gesture) and those that do not is pretty obfuscated...

### DIFF
--- a/fragments/before.frag
+++ b/fragments/before.frag
@@ -1,40 +1,46 @@
 /**
- *  p5.sound extends p5 with <a href="http://caniuse.com/audio-api"
+ *  <p>p5.sound extends p5 with <a href="http://caniuse.com/audio-api"
  *  target="_blank">Web Audio</a> functionality including audio input,
  *  playback, analysis and synthesis.
- *  <br/><br/>
- *  <a href="#/p5.SoundFile"><b>p5.SoundFile</b></a>: Load and play sound files.<br/>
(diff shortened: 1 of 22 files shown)


therewasaguy:
One potential improvement would be to wrap the text like

```
if (getAudioContext().state !== 'running') {
  text('tap to play', width/2, 20);
}
```

Maybe there is a less verbose way...

therewasaguy:
This may be worth discussing (just updated) but I am going to merge and if anybody is inspired to iterate on examples from here, please feel free!!

> So far, some examples use soundFile.play() or oscillator.start() rather than a potentially redundant call to userStartAudio(). It might be worth that redundant call, because the difference between methods that call "start" on a WebAudio node (thus enabling audio on a user gesture) and those that do not is pretty obfuscated...

Issue #417, opened 2020-03-09
no option to set BUFFERSIZE in p5.Amplitude 


endurance21:
there is no option to set bufferSize manually by users while using "amplitude analysis" of a audio .

PR #418, opened 2020-03-09
added feature of manually setting "BUFFERSIZE" of FFT data used in amplitude analysis

endurance21:
solves #417

not only feature is added but also example have been updated ! 

solved first p5-sound js


endurance21:
hey @therewasaguy !
after exploring a lot and playing around with P5.js-sound library ,  i pointed out a new feature and 
have tried to implement the same!
please have  a  look!
thanks!

PR #419, opened 2020-03-09
added feature of manually setting "BUFFERSIZE" in amplitude analysis 

endurance21:
solves #417

not only feature is added but also example have been updated ! 

solved first p5-sound js

areas of impact 
examples/amplitude_analysis/sketch.js
lib/p5.sound.js
lib/p5.sound.js.map
lib/p5.sound.min.js
lib/p5.sound.min.js.map
src/amplitude.js


endurance21:
hey @therewasaguy !
after exploring a lot and playing around with P5.js-sound library ,  i pointed out a new feature and 
have tried to implement the same!
please have  a  look!
thanks!

src/amplitude.js
```diff
@@ -290,6 +290,21 @@ define(function (require) {
...
+   *  @method setBufferSize
+   *  @for p5.Amplitude
+   *  @param {Number} set bufferSize to any value multiple of 2
+   */
+
+  p5.Amplitude.prototype.setBufferSize = function(s) {
+    if (s%2==0) {
+      this._workletNode.port.postMessage({ name: 'bufferSize', bufferSize: s });
+    } else {
+      console.log('Error: bufferSize must be multiple of 2');
```
therewasaguy: are we checking for a multiple of 2, or power of 2? Are there different min/max limits for different browsers?

endurance21:
@therewasaguy
It must be power of 2, thank you for pointing out 😅. I will be correcting it.

examples/amplitude_analysis/sketch.js
```diff
@@ -4,13 +4,15 @@
 
 var soundFile;
 var amplitude;
+var bufferSize = 2048; // a value which can be multiple of 2 , eg. 256 , 512 , 1024 , 2048 ..etc
```
therewasaguy: looks like this comment should be updated to read "power of 2"

src/amplitude.js
```diff
@@ -47,10 +47,10 @@ define(function (require) {
    *
    *  </code></div>
    */
-  p5.Amplitude = function(smoothing) {
+  p5.Amplitude = function(smoothing, bufferSize = 2048) {
 
     // Set to 2048 for now. In future iterations, this should be inherited or parsed from p5sound's default
-    this.bufferSize = safeBufferSize(2048);
+    this.bufferSize = safeBufferSize(bufferSize);
```
therewasaguy: the comment above this line should be deleted or updated. I'm curious if you considered the other options mentioned in the comment?

src/amplitude.js
```diff
@@ -290,6 +290,21 @@ define(function (require) {
...
 
+/**
+   *  bufferSize of Amplitude 
+   *  @method setBufferSize
+   *  @for p5.Amplitude
+   *  @param {Number} set bufferSize to any value multiple of 2
+   */
+
+  p5.Amplitude.prototype.setBufferSize = function(s) {
+    if (s%2==0) {
```
therewasaguy: it's a good practice to use triple equals so that we check for type as well

examples/amplitude_analysis/sketch.js
endurance21: ya ofcourse !!

src/amplitude.js
endurance21: @therewasaguy  
actually comment must be modified  :smile:
 and
as far as inheritance part is concerned i think , i need to modify p5.sound pass a optional argument "bufferSize " like i did in here , but right now let's leave it and i will make it another pr !
what do you say ?

src/amplitude.js
endurance21: actually i have modified it in my local repo , i will be pushing it with your recommendations keeping in mind  :smiley:

endurance21:
@therewasaguy
I have read the w3 specs about web audio api and here is what i get
``
The bufferSize parameter determines the buffer size in units of sample-frames. If it's not passed in, or if the value is 0, then the implementation will choose the best buffer size for the given environment, which will be constant power of 2 throughout the lifetime of the node. Otherwise if the author explicitly specifies the bufferSize, it must be one of the following values: 256, 512, 1024, 2048, 4096, 8192, 16384. This value controls how frequently the audioprocess event is dispatched and how many sample-frames need to be processed each call. Lower values for bufferSize will result in a lower (better) latency. Higher values will be necessary to avoid audio breakup and glitches. It is recommended for authors to not specify this buffer size and allow the implementation to pick a good buffer size to balance between latency and audio quality
 ``
But i think p5 user barely cares about how frequently onaudioprocess task is dispatched and
Neither i find any glitches with or without setting bufferSize
Thus keeping  present requirement

therewasaguy:
hey @endurance21, I've added a card for this to our GSoC backlog. It would need resolving of merge conflicts and the changes mentioned above, but before doing that...it sounds like we're on the same page that this feature isn't as much of a priority right now. How do you feel about closing this so that we can focus?

Issue #411, opened 2020-02-22
Documentation of p5.Oscillator is incomplete


JimBrink:
processing/p5.js#3815 ## Most appropriate sub-area of p5.js?

- [ x] Other (specify if possible)
sound library
#### Details about the bug: 
The documentation for p5.Oscillator is missing some existing trivial methods that may be useful.  Unfortunately, potential users normally can't use them if they do not know they exist.
getFreq()
getAmp()
getType()
getPan()

endurance21:
@therewasaguy 
can you please guide me , how i can get started to add documentations?

therewasaguy:
@endurance21 thanks for your interest! Here's a guide to getting started with documentation: https://github.com/processing/p5.js-sound/wiki/Documentation If anything is unclear, you can update the documentation along the way, and/or ask questions to help clarify.

Essentially you can add documentation by adding JSDoc-style documentation to [this file](https://github.com/processing/p5.js-sound/blob/master/src/oscillator.js), for example https://github.com/processing/p5.js-sound/blob/master/src/oscillator.js#L11

parulpriyedarshani:
@therewasaguy The p5.js Documentation guide link [for](https://github.com/processing/p5.js/blob/master/developer_docs/inline_documentation.md) writing the documentation is giving the 404 error as the Page cannot be found. Could you please help me out with this?

therewasaguy:
thank you, @parulpriyedarshani, and apologies for the delayed response but I've just updated the link in the wiki, and this is the link: https://github.com/processing/p5.js/blob/master/contributor_docs/inline_documentation.md

parulpriyedarshani:
@therewasaguy thank you! That was really helpful.

PR #420, opened 2020-03-09
added documentations for getAmp, getType,getFreq,getPan

endurance21:
solves the issue 
#411
added  documentations for required METHODS
 
areas of impact 
src/oscillator.js


endurance21:
@therewasaguy
first of all sorry, for the late response, as my sem exams were around so I couldn't make the pr soon.

i have added documentations, please see if something need to be added or so!
Thanks

src/oscillator.js
```diff
@@ -309,6 +321,13 @@ define(function (require) {
   p5.Oscillator.prototype.setType = function(type) {
     this.oscillator.type = type;
   };
+  /**
+   *  Returns  current type of oscillator eg. 'sine', 'triangle', 'sawtooth' or 'square'.
+   *
+   *  @method  getType
+   *  @for p5.Oscillator
+   *  @param {String} type 'sine', 'triangle', 'sawtooth' or 'square'.
```
therewasaguy: I think you meant `@returns`? This method doesn't accept a `@param` but those are the possible return values

src/oscillator.js
```diff
@@ -213,6 +213,12 @@ define(function (require) {
   // these are now the same thing
   p5.Oscillator.prototype.fade =   p5.Oscillator.prototype.amp;
 
+  /**
+   * Returns the value of output gain 
+   *
+   *  @method  getAmp
+   *  @for p5.Oscillator
+   */
```
therewasaguy: Should probably add a `@returns {number}` here, the description could be something like `Amplitude value between 0.0 and 1.0`

src/oscillator.js
```diff
@@ -294,6 +300,12 @@ define(function (require) {
       return this.oscillator.frequency;
     }
   };
+   /**
+   * Returns the value of frequency of oscillator 
+   *
+   *  @method  getFreq
+   *  @for p5.Oscillator
+   */
```
therewasaguy: a `@returns` would be helpful here

endurance21:
@therewasaguy 
done and dusted !

src/oscillator.js
```diff
@@ -213,6 +213,14 @@ define(function (require) {
   // these are now the same thing
   p5.Oscillator.prototype.fade =   p5.Oscillator.prototype.amp;
 
+  /**
+   * Returns the value of output gain 
+   *
+   *  @method  getAmp
+   *  @for p5.Oscillator
+   * 
+   * @return  {number} Amplitude value between 0.0 and 1.0
```
therewasaguy: I think we need `@returns` with an "s" at the end :) https://jsdoc.app/tags-returns.html

lib/p5.sound.js
```diff
@@ -1577,6 +1577,14 @@ var __WEBPACK_AMD_DEFINE_RESULT__;
 
 
   p5.Oscillator.prototype.fade = p5.Oscillator.prototype.amp;
+  /**
+   * Returns the value of output gain 
+   *
+   *  @method  getAmp
+   *  @for p5.Oscillator
+   * 
+   * @return  {number} Amplitude value between 0.0 and 1.0
```
therewasaguy: the tag we need is `@returns` with an "s" at the end

src/oscillator.js
endurance21: okay ! i get , my bad

lib/p5.sound.js
endurance21: i think making change in src/oscillator.js is enough , after that this will be build by grunt ?

Issue #422, opened 2020-03-15
getLevel method of p5.soundfFile  is deprecated and still includes in testing 


endurance21:
getLevel method of p5.soundfFile  is deprecated but still includes in testing , it must be removed from testing  to improve our testing !
`
p5.js-sound/src/soundfile.js
`
![getleveleror](https://user-images.githubusercontent.com/43696525/76711780-ae667080-6738-11ea-8a7f-bca07105f7e9.png)

`
p5.js-sound/test/tests/p5.SoundFile.js
`
![testerrorgetlevel](https://user-images.githubusercontent.com/43696525/76711781-b4f4e800-6738-11ea-8fcb-f76b09adc09b.png)

PR #423, opened 2020-03-15
removed the testing of getLevel Method of p5.SoundFile

endurance21:
solves 
#422 

removed the unnessary tests of getLevel Method of p5.SoundFile


endurance21:
@therewasaguy 
please review this!

Issue #413, opened 2020-02-22
gain amp function requires numbers


JimBrink:
This is not an error but clarification would be helpful.  Modification of the code would be even better.
Documentation for p5.Gain's amp function says "Set the output level of the gain node.".  This is unlike the documentation for p5.Oscillator's amp function which clearly states "Set the amplitude between 0 and 1.0. Or, pass in an object such as an oscillator to modulate amplitude with an audio signal."  I expect that first sentence would be appropriate for p5.Gain.  Modifying the code to allow the second would be really nice.

parulpriyedarshani:
@JimBrink  After digging deep into the problem statement, I think this one could be the possible and appropriate clarification for the above:

- As we know, Amplitude is the level of decibels the wave arrives at each cycle without being influenced by gain. A Web Audio sine wave oscillator always produces values between 1 and -1 and all of our signals should fall between this range only. This signal peak amplitude of 1 is constant and cannot be changed. 

- In theory, a sound wave can have any amplitude, and we can boost a wave by multiplying its samples with any arbitrarily large number. But in practice, the Web Audio API limits amplitude to a certain threshold, like here from 0 to 1. The resulting waveform is very different from the original and also sounds completely different if the sound wave doesn't lie in this range.

- Instead, we use a separate gain control to modify the amplitude. Gain is well, how much the wave will be amplified. The main way to affect the loudness of a sound is using GainNodes. The Web Audio API does not use dBSPL(decibels relative to sound pressure level), since the final volume of the sound depends on the OS gain and the speaker gain, and only deals with dBFS(decibels full scale).

-  Again, a Web Audio node called GainNode is used here exactly for this purpose. These nodes have a gain parameter, which acts as a multiplier on the incoming sound buffer. The default gain value is one, which means that the input sound is unaffected. 

- Values between zero and one reduce the loudness, and values greater than one amplify the loudness. Negative gain (values less than zero) inverts the waveform (i.e., the amplitude is flipped). To get a sine wave with half the amplitude of the default, we can connect an OscillatorNode to a GainNode whose gain parameter is set to 0.5(say). The result is a sine wave that oscillates between -1 * 0.5 = -0.5 and 1 * 0.5 = 0.5. What we have built here is a Web Audio graph with three nodes in it. The audio signal is generated by the oscillator, and flows through the gain before reaching the destination. The gain alters the signal by halving its amplitude.

[...]

PR #426, opened 2020-03-17
Modification added for the clarification of Gain.js amp() function

parulpriyedarshani:
Solved Issue #413 
Added documentations for the clarification of the amp() function of p5.gain component.

### DIFF
--- a/src/gain.js
+++ b/src/gain.js
@@ -124,8 +124,16 @@ define(function (require) {
   };
 
   /**
-   *  Set the output level of the gain node.
-   *
+   *  The main way to affect the loudness of a sound is using a Web Audio API node called GainNode. GainNode always has exactly one input and one output, both with the same number of channels.
+   *  It is an AudioNode audio-processing module that causes a given gain to be applied to the input data before its propagation to the output.
+   *  Sets the output level of the GainNode. These nodes have a gain parameter, which acts as a multiplier on the incoming sound buffer. Gain is the amount of amplification done to the signal.
+   *  The gain is a unitless value, changing with time, that is multiplied to each corresponding sample of all input channels. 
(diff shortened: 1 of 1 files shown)


parulpriyedarshani:
@therewasaguy Please, kindly review the PR.

parulpriyedarshani:
@therewasaguy Please, kindly review this PR.

endurance21:
hey ! @parulpriyedarshani 
your work seems very nice to me!
i was asking  can we add a little more info like ,
 " The new gain is instantly applied, causing unaesthetic 'clicks' in the resulting audio .  
In Case you need to prevent this from happening, you can use " rampTime " and " tFromNow " arguments to add a fading effect to the sound "

src/gain.js
```diff
@@ -124,8 +124,13 @@ define(function (require) {
   };
 
   /**
-   *  Set the output level of the gain node.
-   *
+   *  Set the output level of the GainNodes.Gain is the amount of amplification done to the signal. 
+   *  The main way to affect the loudness of a sound is using a Web Audio node called GainNodes.
```
therewasaguy: Please clarify that it's the Web Audio API `GainNode` (no "s" at the end). If mentioning the underlying Web Audio API components, it could be nice to link to their documentation on MDN.

parulpriyedarshani:
@therewasaguy Thanks for your review! :+1: For the first issue, yes, it will be great to unify all the documentation related to all the .amp methods for which we can open a particular issue.

src/gain.js
parulpriyedarshani: Could you please explain a little bit what did you actually mean by "linking to their documentation on MDN ".

parulpriyedarshani: Resolved :+1:

PR #428, opened 2020-03-17
Added exponentialRampToValueAtTime feature

### LINKED ISSUE #427: Adding exponentialRampToValueAtTime() to GainNodes
while Setting the value of AudioParams such GainNode.gain, the net gain is instantly applied, causing unaesthetic 'clicks' in the resulting audio. To prevent this from happening,
we should never change the value directly but use the exponential interpolation methods on the AudioParam interface.
Also fading effect is very common nowAdays in Modern Audio Players !
Right now we support have for linearRampToValueAtTime() , 
exponential

i am proposing to add an extra optional argument to  p5.Gain.prototype.amp Method to allow Exponential fade effect ! 
like this 
solveExp

endurance21:
fixes 
#427

### DIFF
--- a/src/gain.js
+++ b/src/gain.js
@@ -130,17 +130,21 @@ define(function (require) {
    *  @for p5.Gain
    *  @param  {Number} volume amplitude between 0 and 1.0
    *  @param  {Number} [rampTime] create a fade that lasts rampTime
-   *  @param  {Number} [timeFromNow] schedule this event to happen
-   *                                seconds from now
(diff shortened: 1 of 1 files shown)


endurance21:
@therewasaguy
Please review the Pr

endurance21:
I will be adding unit tests for this soon 😊

therewasaguy:
@endurance21 how would you feel about adding this idea to the backlog for now?

PR #440, opened 2020-03-27
added feature of safeBins in fft method of P5 along with unit Test

endurance21:
issue :
safebins

the bins arguments is very vulnerable , as there is no check to what user enters the input , suppose the user enters a negative value of bins by mistake , then it's a crash !
so there must be a check to what user gives input as  "bins" arguments"

solution : 
* added a filter (safeBins) to what user enters as argument .
this bins

* checks are esablished as shown 
safeBins
* added a unit test for the function
testSafeBins


endurance21:
@therewasaguy 
 please  review this Pr!

src/helpers.js
```diff
@@ -337,12 +337,27 @@ define(function (require) {
 
     return bufferSize;
   }
+ var safeBins =  p5.prototype.safeBins  = function(bins) {
+    let safeBins = 1024;
+    if(typeof(bins)==="string"){
+      console.log("the value of bins must be power of two and between 16 and 1024");
```
therewasaguy: should this be a `console.warn` or `console.error` ?

src/helpers.js
```diff
@@ -337,12 +337,27 @@ define(function (require) {
 
     return bufferSize;
   }
+ var safeBins =  p5.prototype.safeBins  = function(bins) {
```
therewasaguy: do we need to add this to the p5 prototype, or can this be a private method that is only used internally?

If it's only added to the prototype for the purpose of writing a test, I think the test should instead be on the public methods that utilize this function behind the scenes.

test/tests/utils/test.helpers.js
```diff
@@ -0,0 +1,35 @@
...
+            let test = 1500;
+            expect(safeBins(test)).to.equal(1024);
+        });
+        it('can handle value equal to 0',()=>{
+            let test = 0;
+            expect(safeBins(test)).to.equal(1024);
+        });
+        it('can handle strings',()=>{
+            let test = 'testString';    
+            expect(safeBins(test)).to.equal(1024);
```
therewasaguy: all of these tests "handle" _invalid_ values by setting a default value.

Other libraries might choose to throw an error when a user enters an invalid value. Since we're not throwing an error, I wonder if there's a way we could verify that an error message was logged to the console?


It is useful to test behavior when a user provides invalid values. But we probably also want to test for valid value(s). For example, with the tests so far, the method could be changed to _always_ return 1024 and we wouldn't realize this bug.

test/tests/utils/test.helpers.js
```diff
@@ -0,0 +1,35 @@
...
+        });
+
+    });
+
+});
```
therewasaguy: this 🚫 means that the file doesn't end with a new line. I don't think it's a big deal but some do, so I think it's a best practice to end files with a new line :) https://thoughtbot.com/blog/no-newline-at-end-of-file

test/tests/utils/test.helpers.js
endurance21: ya ! this is the bug , if anyone changes the function to always return 1024 then our also this test will pass . okay i will be including that

endurance21:
> Also, just a note:
> 
> In the Web Audio API, the `AnalyserNode.frequencyBinCount` is `Read only` (I'm looking at the MDN documentation for reference https://developer.mozilla.org/en-US/docs/Web/API/AnalyserNode). It is always half of the fftSize. So we're kind of doing things backwards in p5.Sound by allowing users to set the binSize directly. We're doing that because it seemed like it would be more intuitive to set a value that represents the array size you get from the analysis, rather than a value that is twice that amount... but I'm having second thoughts about that now.


yeah ! you are right ! but it is very intuitive and generous to provide user the flexibilty  to choose length of array in which analysed data will be served to them !, and we are no where assigning the value to "AnalyserNode.frequencyBinCount" that means it is good to use ,
infact we are creating a level of abstraction that provides ease to user !, so we don't need to remove that ,however we can rename it , name other than bins because it seem to  represent "AnalyserNode.frequencyBinCount" 
how about name "length" which represents the length of array !

endurance21:
> I suggest testing the public methods, and making this a private method.
i was actually thinking the same , means i need to create a wrapper for this method ? or you have any other  suggestions?

endurance21:
> Do you think the functionality of `safeBins` could be incorporated into `safeBufferSize`, rather than making it a separate method?


we can create a safeCheck function which can be placed inside of both of these methods and 
rather writing the same check twice , also safeBufferSize have other checks than  safeBIns has
image
Also , 
https://github.com/processing/p5.js-sound/pull/419#issuecomment-603586872 
 i agree to the w3 specs and not giving our user the ability to change the bufferSIZe() and hence we don't need to add tests for it like in safeBins !!

src/helpers.js
endurance21: a warning would be forgiving and error will remind user of its mistake ,
but here as even after the user enters wrong value we assign it a default one so it think , we are not breaking the operation so i think giving warning will be legitimate console.warn ()

src/helpers.js
endurance21: @therewasaguy  the main problem i face was using AMD here , I was unable to use safeBins defined in src/helper.js in test/tests/helpers.test.js , so I decided to put this method in prototype object of p5 ,
and I am realising now , that it is not the good way of doing it !,
so  I will add tests on   public function that will wrap this function ! and remove it from p5.prototype

test/tests/utils/test.helpers.js
endurance21: are you talking about "\n" or what ?
 
i don't see this newline  in any of the test file in our library! should I use it everywhere ?

> this  means that the file doesn't end with a new line. I don't think it's a big deal but some do, so I think it's a best practice to end files with a new line :) https://thoughtbot.com/blog/no-newline-at-end-of-file

endurance21:
i would like to draft this one until the module system work is done !

Issue #313, opened 2018-08-05
SoundLoop interval of 0


JunShern:
Hi,

I've found that whenever I set

```
mySoundLoop.interval = 0 
```

for a p5.SoundLoop, the sketch freezes (presumably because it tries to run the SoundLoop infinitely quickly). However, in the case of updating the SoundLoop based on note event deltatimes, it is often useful to have an interval of 0.

The quick fix I'm using now is to just do

```
mySoundLoop.interval = max(desiredInterval, 0.01) 
```

But we might want to add this into the source code itself. Any thoughts? @therewasaguy 

Thanks!

parulpriyedarshani:
@JunShern I agree with your solution and would like to work on this issue. After doing some research I was unable to find desiredInterval in the codebase. Could you please tell me about how this desiredInterval is working and where has it been initialized in the code?
Thanks

PR #458, opened 2020-04-28
Loop

parulpriyedarshani:
Solves issue #313 
Initialized the desiredInterval to zero and set the soundLoop interval of 0.


parulpriyedarshani:

src/soundLoop.js
```diff
@@ -65,6 +65,12 @@ define(function (require) {
 
     this._interval = interval || 1;
 
+    // Set the limit of the desiredInterval to zero.
+
+    let desiredInterval = 0;
+
+    SoundLoop._interval = max(desiredInterval, 0.01);
```
therewasaguy, line 72: hrm, `SoundLoop` doesn't exist, instead we have a function called p5.SoundLoop that exists on the p5 object.

We don't want to set _interval on the class, but not on the instance (which is represented by `this`).

That is happening on line 66. So we should do this check on that line.

src/soundLoop.js
```diff
@@ -65,6 +65,12 @@ define(function (require) {
 
     this._interval = interval || 1;
```
therewasaguy, line 66: can we do the check on this line? `this._interval = Math.max(interval, 0.01)` ? And then, the question is how can we set the default? We could set the default in the constructor function on line 57, by saying `interval = 1` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

Abhijay007 (commented):
@parulpriyedarshani the code is iterated a lot in the past year, I am not sure if the soundLoop.interval problem still exists or not, can you please look into this once? due to branch change it has some merge conflicts too that need to be addressed

Issue #374, opened 2019-07-10
buggy behaviour in part.onStep()


i3games:
the pattern below has 8 beats, play() is called 8 times but `part.onStep` is called 7 times. 
When the part is looping, `part.onStep` is still only called 7 times.

```javascript

let sound, m = 0, n = 0;

function setup () {
  masterVolume(0.4);

  const pat = [1,1,1,1,1,1,1,1];
  const phrase = new p5.Phrase('phrase', play, pat);

  part = new p5.Part();
  part.onStep(() => {
    console.log(`step ${++n}`);
  });
  part.addPhrase(phrase);
  part.setBPM(60);
  part.start();
  // part.loop();
  
  sound = new p5.MonoSynth();

  userStartAudio().then(() => {
     console.log("user start audio");
   });
}

function draw () {
  background(0);
}

function play (time, notused) {
  console.log(`play ${++m}`);
  sound.play('D4');
}

```

parulpriyedarshani:
I have got a solution for this, I would like to work on this issue.

parulpriyedarshani:
I am working on this issue but I'm unable to find this particular code snippet. Could anyone please help me in finding this particular code snippet in the code base?

i3games:
Output:

```
user start audio sketch.js:21:14
step 1 sketch.js:11:13
play 1 sketch.js:30:11
step 2 sketch.js:11:13
play 2 sketch.js:30:11
step 3 sketch.js:11:13
play 3 sketch.js:30:11
step 4 sketch.js:11:13
play 4 sketch.js:30:11
step 5 sketch.js:11:13
play 5 sketch.js:30:11
step 6 sketch.js:11:13
play 6 sketch.js:30:11
step 7 sketch.js:11:13
play 7 sketch.js:30:11
done p5.sound.js:10794:17
play 8 sketch.js:30:11
```

parulpriyedarshani:
@i3games Thanks for posting the output. I am unable to find this particular code snippet in the entire codebase. I was wondering if you could just help me out with the file in which this particular snippet exists. That would be really very helpful.

i3games:
Hi, the above code was my program demonstrating the problem.

i3games:
I just checked with p5.js 1.0.0 and the problem still persists.

parulpriyedarshani:
@i3games Thank you, It seems good to me using part.loop() instead of part.start(). It worked for me. Well, I'm working on this issue and have opened a PR for the same.

PR #459, opened 2020-04-29
Fixed buggy behaviour in part.onStep()

parulpriyedarshani:
Fixed issue #374 
Used part.loop() instead of part.start() to call part.onStep() 8 times for each loop as long as it is looping.


parulpriyedarshani:

therewasaguy:
I don't think this fixes the issue, since the modification was in the /examples directory rather than in the /src directory

Issue #76, opened 2015-11-07
Some Oscillator methods may not make sense with p5.Noise


toolness:
I'm a total audio noob and was just fiddling around with the library and noticed that calling the `freq()` method on an instance of `p5.Noise` raises the following exception:

```
TypeError: this.oscillator is undefined
```

This happened because I had some code that instantiated a `p5.Oscillator` and did some things with it. When I read the [p5.Noise documentation](http://p5js.org/reference/#/p5.Noise) stating that "Noise is a type of oscillator", I figured that means I should be able to replace my Oscillator instance with a Noise instance and see (er, hear) something cool happen.

Only after further introspection did I realize that the notion of setting a "frequency" for "noise" is a weird idea, so perhaps the `freq()` method isn't meant to be used by Noise instances in the first place. However, in the spirit of p5's [friendly error system](https://github.com/processing/p5.js/issues/971), it seems like overriding `freq()` on the Noise class and making it throw an exception with a more helpful message might be more useful to beginners.

parulpriyedarshani:
@b2renger Are you still working on this issue. Can I proceed with this issue.

parulpriyedarshani:
@b2renger As there has been no activity on this issue for a very long time and I had got a solution for the issue, so I have added a PR fixing the same.

PR #461, opened 2020-04-30
Added demonstration for freq() method in p5.Noise

parulpriyedarshani:
Fixed issue #76 
Added demonstration for overriding the freq() method on the Noise class and making it throw an exception with a more helpful message that would be useful to beginners.

### DIFF
--- a/src/noise.js
+++ b/src/noise.js
@@ -79,7 +79,41 @@ define(function (require) {
     brownBuffer.type = 'brown';
     return brownBuffer;
   })();
+  /**
+   * In the spirit of p5’s friendly error system, this is clearly an example for overriding the freq() on the Noise class 
+   * by calling a Noise instance (as “Noise is a type of oscillator”) and make it throw an exception,
+   * so that it proves to be useful for beginners. 
+   * The freq() method isn’t meant to be used by Noise instances and hence, will throw an exception.
+   * 
+   * Set frequency of an oscillator to a value. Or, pass in an object
+   * such as an oscillator to modulate the frequency with an audio signal. 
+   * 
+   * Set frequency of an oscillator to a value. Or, pass in an object
+   *  such as an oscillator to modulate the frequency with an audio signal.
+   *
+   *  @method  freq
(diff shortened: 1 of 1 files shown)


parulpriyedarshani:

src/noise.js
```diff
@@ -79,7 +79,41 @@ define(function (require) {
...
+   *  @param  {Number|Object} Frequency Frequency in Hz
+   *                                        or modulating signal/oscillator
+   *  @param  {Number} [rampTime] Ramp time (in seconds)
+   *  @param  {Number} [timeFromNow] Schedule this event to happen
+   *                                   at x seconds from now
+   *  @return  {AudioParam} Frequency If no value is provided,
+   *                                  returns the Web Audio API
+   *                                  AudioParam that controls
+   *                                  this oscillator's frequency
+   */
```
therewasaguy, line 105: let's not add documentation for this method. If we do, it will show up on the website, and that might confuse people since the point is that they should not be calling this method.

src/noise.js
```diff
@@ -79,7 +79,41 @@ define(function (require) {
...
 
+  p5.Noise.prototype.freq = function(val, rampTime, tFromNow){
+    // passing values for the parameters.
+    var example = new p5.Noise(440, 0.1, 1);
```
therewasaguy, line 109: I'm not sure why we would want to create a new instance of p5.Noise inside of this method?

src/noise.js
```diff
@@ -79,7 +79,41 @@ define(function (require) {
...
+    // calling the freq() method
+    example.freq(()=>{
+      if(!this.output){
+        console.error('TypeError:value not of the expected type.`freq()` is not available for p5.Noise class. Can’t be used for controlling the noise of an audio signal as that would be absurd.');
```
therewasaguy, line 113: I think this method could simply say 'console.error(`freq()` is not available for p5.Noise class)'

therewasaguy:
closing for now, thank you for the PR

PR #462, opened 2020-04-30
added unit test for p5.gain in es6

endurance21:
added unit tests for all the methods in p5.Gain 
here are the test  results and AUDIO GRAPH ( inspected with WEB AUDIO INSPECTOR) of this unit test 
Selection_086

### DIFF
--- a/test/tests/p5.Gain.js
+++ b/test/tests/p5.Gain.js
@@ -19,14 +19,16 @@ define(['chai'],(chai)=>{
             
             mainGainNode.connect(outputNode);
         });
-        it('can disconnect  all output' ,()=>{
-            let inputNode = new p5.Gain();
-            let outputNode = new p5.Gain();
-            let mainGainNode  = new p5.Gain();
+        it('can disconnect  from the given output if present any' ,()=>{
+            let outputNode1 = new p5.Gain();
+            let mainGainNode1  = new p5.Gain();
 
-            mainGainNode.setInput(inputNode);
-            mainGainNode.connect(outputNode);
-            mainGainNode.disconnect();
+            mainGainNode1.connect(outputNode1);
+            mainGainNode1.disconnect();         // it disconnects from the output if present 
+
+            let mainGainNode2 = new p5.Gain();
+            mainGainNode2.disconnect();         // the disconnects can handle things if there is no output too 
+
         });
         it('can set the output level of gain Node', ()=>{
             let osc = new p5.Oscillator('sine');
--- a/test/tests/p5.Gain.js
+++ b/test/tests/p5.Gain.js
@@ -1,0 +1,51 @@
+'use strict '
+
+define(['chai'],(chai)=>{
+    const expect = chai.expect ;
+    describe('P5.Gain',()=>{
+        it('can be initilized and disposed',()=>{
+            let gain = new p5.Gain();
+            gain.dispose();
+        });
+        it('can set input', ()=>{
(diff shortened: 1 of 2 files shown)


endurance21:
@therewasaguy  
please review this PR ! 
Also, i have been using this amazing `WEB AUDIO INSPECTOR`  tool to see the AUDIO GRAPHS 
, I find it very useful while working with web audio API, i am curious if we can use this as integrated part of our codebase, it will really help the developers to understand and write codes ,specially when they can see what they are creating!

endurance21:
@therewasaguy we can add it too?

test/tests/p5.Gain.js
```diff
@@ -0,0 +1,51 @@
...
+            let gain = new p5.Gain();
+            gain.dispose();
+        });
+        it('can set input', ()=>{
+            let inputNode = new p5.Gain();
+            let mainGainNode  = new p5.Gain();
+
+            mainGainNode.setInput(inputNode);
+        });
+        it('can send output',()=>{
```
therewasaguy, line 16: this test is doing the same thing as 'can set input' maybe they can just be one test?

test/tests/p5.Gain.js
```diff
@@ -0,0 +1,51 @@
...
+            mainGainNode.connect(outputNode);
+        });
+        it('can disconnect  all output' ,()=>{
+            let inputNode = new p5.Gain();
+            let outputNode = new p5.Gain();
+            let mainGainNode  = new p5.Gain();
+
+            mainGainNode.setInput(inputNode);
+            mainGainNode.connect(outputNode);
+            mainGainNode.disconnect();
```
therewasaguy, line 29: this is testing that we are able to call the `disconnect` method without that throwing any runtime errors, but how do we know if we successfully disconnected? (I'm not sure!)

Also, will this disconnect the inputNode or only the connection between the mainGaneNode and the outputNode? I think it's only that one. Again, how would we know from this test?

test/tests/p5.Gain.js
endurance21: though they appear to do the same task but actually we are able to validate twoo methods here 
setInput() and Connect() method

therewasaguy: ahh, got it! Thank you for clarifying and sorry I missed this

Abhijay007:
Hi @endurance21,  can you please look into the merge conflicts, if you are busy I can resolve those conflicts and make a new PR around the same :)

Abhijay007 (commented):
Seems like the Test for p5.Gain.js is already updated and present in the repo, refer: https://github.com/processing/p5.js-sound/blob/main/test/tests/p5.Gain.js (last updated by @satyasaibhushan,  

@davepagurek I think we can close this PR now.

davepagurek:
We may want to copy over just the `can set the output level of gain Node` test, since the current test file doesn't test changing the gain at all. But I think that's the only one that isn't covered now!

Abhijay007:
> We may want to copy over just the `can set the output level of gain Node` test, since the current test file doesn't test changing the gain at all. But I think that's the only one that isn't covered now!

okay got it, so can I open a PR to add the `can set the output level of gain Node`  in the existing gain test?

davepagurek:
yeah sounds good!

Abhijay007:
Okay will do !!

davepagurek:
Thanks @Abhijay007! I'll close this PR and carry over reviews on your new one.

PR #463, opened 2020-05-01
added unit test for p5.master

endurance21:
added unit test for the methods in p5.master 

1.checks for value mastervolume ( set and returned values)
2.checks  for  fading effect of setting masterVolume 
3.checks for setting the masterVolume after the given time t 
results 

Selection_087

@therewasaguy please review this pr!

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -8,6 +8,7 @@ require.config({
 });
 
 var allTests = [
+  'tests/p5.Master',
   'tests/p5.SoundFile',
   'tests/p5.Amplitude',
   'tests/p5.Oscillator',
--- a/test/tests/p5.Master.js
+++ b/test/tests/p5.Master.js
@@ -1,0 +1,37 @@
+'use strict'
+
+define(['chai'], function(chai){
+    const expect = chai.expect;
+    describe('P5.master',function(){
+
+        it('can set and return master volume',function(done){
+            p5.prototype.masterVolume(.6);
+
+            setTimeout(function() {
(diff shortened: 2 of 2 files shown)


test/test.js
```diff
@@ -8,6 +8,7 @@ require.config({
 });
 
 var allTests = [
```
montoyamoraga, line 10: does it need to be var? shouldnt' we use ES6, let?

endurance21: @montoyamoraga 
yes , we should  !
in fact i have just checked and it works fine as it should be !
:grin:

montoyamoraga: @endurance21 can you include that in this PR? i only know that unit testing exists, but i don't know what they are for, or how to check them, i am assigning this PR to somebody else for review :)

endurance21: @montoyamoraga 
however we must also consider using `const` here , as the reference is not changing ayway !

endurance21: > @endurance21 can you include that in this PR? i only know that unit testing exists, but i don't know what they are for, or how to check them, i am assigning this PR to somebody else for review :)

for sure i will be adding it !

endurance21: but i think i have to make another pr for that , otherwise there might  be merge conflicts with other PRs !  :confused:
i will make another pr for that once other things  of this PR  get reviewed and ready to go !
:grinning:

endurance21:
@therewasaguy  i think we can add it  ?

Issue #304, opened 2018-06-26
monosynth.triggerRelease() does not reach 0


therewasaguy:
MonoSynth is using Envelope's `ramp` method, which uses `setTargetAtTime`, and that never reaches the target value.

parulpriyedarshani:
@therewasaguy After going through this issue I think that directly passing the third parameter i.e 0 in the ramp method is causing problems. In my opinion, by initializing a new parameter such as a targetValue to 0 and passing it through the Envelope's ramp method so that it can reach the target of 0 will help in solving the above problem. I have opened a PR, implementing the same. Let me know if this fixes the issue.
Thanks

PR #466, opened 2020-05-01
monoSynth.triggerRelease() reaches the target value

parulpriyedarshani:
Fixed issue #304 
Initialized the targetValue of ramp function for monoSynth.triggerRelease() so it reaches the target value of 0.

### DIFF
--- a/src/monosynth.js
+++ b/src/monosynth.js
@@ -200,9 +200,10 @@ define(function (require) {
      *  }
      *  </code></div>
      */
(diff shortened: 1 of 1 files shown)


parulpriyedarshani:

therewasaguy:
thanks for the PR @parulpriyedarshani. I don't think we need to add this parameter to triggerRelease method, it can always release to 0. If a user wants to ramp to a specific value that isn't zero, they can use triggerAttack. Does that sound reasonable to you?

PR #470, opened 2020-05-02
removed initilized but never used variables from listner3d.js

endurance21:
image
`Effect` variable was initilized but never used ! 
so removed it !!


endurance21:
@therewasaguy  @montoyamoraga 
please review this pr!

PR #502, opened 2020-07-15
replacement of es5 functions to es6 class def feat.P5.FFT

endurance21:
area covered 
src/fft.js


src/fft.js
```diff
@@ -139,7 +141,7 @@ p5.FFT = function (smoothing, bins) {
  *  @for p5.FFT
  *  @param {Object} [source] p5.sound object (or web audio API source node)
  */
-p5.FFT.prototype.setInput = function (source) {
+FFT.prototype.setInput = function (source) {
```
therewasaguy, line 144: rather than use .prototype, we can place these methods within the class, similar to the `constructor` method (which should be at the top), like

```
class FFT {
...
  setInput(source) {
   ...
  }
}
```

therewasaguy:
Looks to be off to a good start! Just a small comment

src/fft.js
endurance21: and yes for sure , we can do so , but a Big BUT is that , 
declaring the method this way , significantly helps in memory optimizations ! 
as we can see here 
https://www.thecodeship.com/web-development/methods-within-constructor-vs-prototype-in-javascript/

should we consider those memory optimization techniques ! ?

kyle1james: Great point to bring up and I'm looking forward to hearing more! Do the con's of memory outweigh the pros of having access to vars inside the constructor? 

https://stackoverflow.com/questions/4508313/advantages-of-using-prototype-vs-defining-methods-straight-in-the-constructor

therewasaguy: These links are from before ES6.

It's true that we don't want to define the method _within_ the constructor.

To clarify: I'm suggesting that we define it _next to / after_ the constructor, as part of the class. This is syntactic sugar for adding the method to the prototype.

```
class FFT {
  constructor() { ... }
  setInput(source) { ... }
}
```

endurance21: @therewasaguy  wow! i just confirmed it , i made a class and tested it .prototype and its object  __proto__ and wolla !
i saw , method was  present in there in its prototype object ! 
That was a nice concept !  i

kyle1james: That syntactical sugar will get you every time.

endurance21: @kyle1james  ohk ohk

endurance21:
@therewasaguy @kyle1james require your review !

endurance21:
Yes that is for later prs 😇

src/fft.js
```diff
@@ -84,534 +84,542 @@ import p5sound from './master';
...
-      get: function () {
-        return this.analyser.fftSize / 2;
-      },
-      set: function (b) {
-        this.analyser.fftSize = b * 2;
+class FFT {
+  constructor(smoothing, bins) {
+    this.input = this.analyser = p5sound.audiocontext.createAnalyser();
+
+    Object.defineProperties(this, {
```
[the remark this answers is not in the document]
endurance21: Okay i will keep that in separate pr

PR #503, opened 2020-07-25
replacement of es5 functions to es6 class def feat .P5.Amplitude 

endurance21:
interface covered


src/audioin.js
```diff
@@ -84,7 +85,7 @@ p5.AudioIn = function (errorCallback) {
    *
    *  @property {p5.Amplitude} amplitude
    */
-  this.amplitude = new p5.Amplitude();
```
therewasaguy, line 87: good call here 👍

PR #508, opened 2020-08-05
 replacement of es5 functions to es6 class def feat P5.AudioIn and audiocontext

endurance21:
## area covered 
* src/audioIn
* src/audiocontext
* src/app

PR #509, opened 2020-08-05
replacement of es5 functions to es6 class def feat P5.AudioVoice , P5.monosynth , p5.Envelope

endurance21:
## area covered
* src/AudioVoice 
* src/Monosynth


endurance21:
this PR shall merge after a successful merge of PR#508

src/app.js
```diff
@@ -33,9 +40,14 @@ import './compressor';
 import './soundRecorder';
 import './peakDetect';
 import './gain';
-import './monosynth';
-import './polysynth';
+// import './monosynth';
+// import './polysynth';
```
therewasaguy, line 44: let's remove the code rather than comment it out

src/app.js
```diff
@@ -1,6 +1,10 @@
 import 'audioworklet-polyfill';
 import './shims';
-import './audiocontext';
+
+import { getAudioContext, userStartAudio } from './audiocontext';
+p5.prototype.getAudioContext = getAudioContext;
+p5.prototype.userStartAudio = userStartAudio;
```
therewasaguy, line 6: looks like this includes changes from #508, best to keep the changes in separate PR's if possible (or put them all in the same PR)

src/oscillator.js
```diff
@@ -626,4 +626,5 @@ p5.SqrOsc = function (freq) {
 
 p5.SqrOsc.prototype = Object.create(p5.Oscillator.prototype);
 
-export const Oscillator = p5.Oscillator;
+const Oscillator = p5.Oscillator;
+export default Oscillator;
```
therewasaguy, line 630: is changing from a named export (that can be imported like `import { Oscillator } from ...`) to a default export (that is imported like `import OscillatorOrWhateverWeWantToCallIt from ...`) going to cause any issues?

src/envelope.js
```diff
@@ -897,4 +897,5 @@ p5.Env = function (t1, l1, t2, l2, t3, l3) {
 };
 p5.Env.prototype = Object.create(p5.Envelope.prototype);
 
-export default p5.Envelope;
+const Envelope = p5.Envelope;
+export default Envelope;
```
therewasaguy, line 901: ah, this was already a default export and we just want to import it by name, this makes sense!

src/oscillator.js
endurance21: @therewasaguy i dont think so, its just for this pr., when i will be changing oscillator file i will make it simply 
`export default oscillator`

We can bear it for now 😉

src/app.js
endurance21: Okayy

src/app.js
endurance21: Thats why i kept it for review after 508 😅

endurance21:
@therewasaguy  please review pr #508  then we can merge this one too !

src/app.js
therewasaguy: un-resolving this because it doesn't appear to be resolved yet

src/oscillator.js
therewasaguy: ah, ok! I did some more digging and it looks like you're right, this won't cause any issues just yet.

The way you've imported Oscillator in MonoSynth looks good https://github.com/processing/p5.js-sound/pull/509/files#diff-559e243463d4dac174c5eee32f469a81R4

For a future PR, iit looks like Pulse should be importing the module as well in order to extend the class.

https://github.com/processing/p5.js-sound/blob/d30c43d9c688f147a5c2ca678da5fd4f1fe24817/src/pulse.js#L49

src/app.js
```diff
@@ -33,9 +40,14 @@ import './compressor';
 import './soundRecorder';
 import './peakDetect';
 import './gain';
-import './monosynth';
-import './polysynth';
+// import './monosynth';
```
therewasaguy, line 43:
```suggestion

```

src/app.js
therewasaguy, line 44:
```suggestion
```

endurance21: Sorry for that, let me make this comment out

endurance21:
@therewasaguy  i have removed the comments! 
good to go?
also i think i must removed the older redudandant commits here ! 
however i think i must squash and commit ? instead of removing the older commits?

therewasaguy:
yes, feel free to "Squash and merge"

looks like there's no conflict with the older commits :)

Screen Shot 2020-08-10 at 10 07 58 AM

PR #511, opened 2020-08-10
requiring user interaction before running tests, per chrome autoplay policy

aceslowman:
This resolves some issues I was having running the tests in Chrome

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -27,6 +27,23 @@ var allTests = [
 
 p5.prototype.masterVolume(0);
 
+var test_has_run = false;
+
 require(allTests, function(){
-  mocha.run();
+  document.getElementById('mocha').innerHTML = 'click to begin tests';
+
+  // chromes autoplay policy requires a user interaction
+  // before the audiocontext can activate
+  function mousePressed() {
+    if(!test_has_run) {
+      document.getElementById('mocha').innerHTML = '';
+      getAudioContext().resume()
+      mocha.run();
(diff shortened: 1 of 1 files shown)


test/test.js
```diff
@@ -27,6 +27,23 @@ var allTests = [
...
 require(allTests, function(){
-  mocha.run();
+  document.getElementById('mocha').innerHTML = 'click to begin tests';
+
+  // chromes autoplay policy requires a user interaction
+  // before the audiocontext can activate
+  function mousePressed() {
+    if(!test_has_run) {
+      document.getElementById('mocha').innerHTML = '';
+      getAudioContext().resume()
```
therewasaguy, line 40: Can we use https://p5js.org/reference/#/p5/userStartAudio (in this case, `p5.prototype.userStartAudio`) for this?

aceslowman: This works! Just pushed new commit. When requiring p5, what is the best way to include the p5.js library? I was running into some issues before realizing that libs/p5.js was missing, which was resolved after manually importing the file, but I don't know if I saw anything about that in the contributions docs.

endurance21:
@aceslowman would you like to merge this one!!
Go ahead

aceslowman:
@endurance21 Great! I don't think I have permissions to merge though, or at least I'm not seeing the 'Merge Pull Request' button.

endurance21:
Oh yes, so let me do this work🤫

PR #523, opened 2020-08-15
Replacement of es5 functions to es6 classes feat p5.polySynth

endurance21:
@therewasaguy @kyle1james Kindly review this PR.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -6,7 +6,10 @@ p5.prototype.getAudioContext = getAudioContext;
 p5.prototype.userStartAudio = userStartAudio;
 
 import './master';
-import './helpers';
+
+import { freqToMidi } from './helpers';
+p5.prototype.freqToMidi = freqToMidi;
+
 import './errorHandler';
 import './audioWorklet';
 import './panner';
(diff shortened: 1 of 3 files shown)


src/app.js
```diff
@@ -6,7 +6,10 @@ p5.prototype.getAudioContext = getAudioContext;
 p5.prototype.userStartAudio = userStartAudio;
 
 import './master';
-import './helpers';
+
+import { freqToMidi } from './helpers';
+p5.prototype.freqToMidi = freqToMidi;
```
therewasaguy, line 11: I'm confused why this change is happening in this PR? There are a bunch of methods that get attached to the p5.prototype as part of helpers, let's be sure to add them all (when `helpers` is addressed)

src/helpers.js
```diff
@@ -27,7 +27,7 @@ p5.prototype.sampleRate = function () {
  *                             above Middle C is 440Hz
  *  @return {Number}   MIDI note value
  */
-p5.prototype.freqToMidi = function (f) {
+export const freqToMidi = function (f) {
```
therewasaguy, line 30: ah, I see how why this change was made 👍

PR #524, opened 2020-08-16
Replacement of es5 functions to es6 classes feat p5.listener3D

endurance21:
@therewasaguy @kyle1james Please review this PR.


src/listener3d.js
```diff
@@ -24,245 +24,239 @@ import p5sound from './master';
 //   * @param {String} [type] [Specify construction of a spatial panner or listener]
 //   */
```
therewasaguy, line 25: this class isn't in the public documentation, but good to update it anyway!

PR #526, opened 2020-08-16
replacement of es5 functons to es6 classes feat p5.filter

endurance21:
@therewasaguy  @kyle1james here we go with one more in a row !

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -30,7 +30,7 @@ p5.AudioIn = AudioIn;
 import Effect from './effect';
 p5.Effect = Effect;
 
-import { Filter, LowPass, HighPass, BandPass } from './filter';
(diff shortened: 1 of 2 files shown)


src/app.js
```diff
@@ -27,10 +27,14 @@ import './noise';
 import AudioIn from './audioin';
 p5.AudioIn = AudioIn;
 
-import Effect from './effect' ;
+import Effect from './effect';
 p5.Effect = Effect;
 
-import './filter';
+import { Filter, LowPass, HighPass, BandPass } from './filter';
```
therewasaguy, line 33: I think Filter should be the default export, just to stay consistent with other modules where the filename is the name of the default export

endurance21: done the required changes ! have a look!

PR #531, opened 2020-08-20
replacement of es5 functions  to es6 classes feat p5.soundFile

endurance21:
all test working fine ! 

@therewasaguy  @kyle1james please review this pr! 

Screenshot from 2020-08-20 19-01-22

### DIFF
--- a/src/soundfile.js
+++ b/src/soundfile.js
@@ -2,1323 +2,9 @@ import CustomError from './errorHandler';
 import p5sound from './master';
 import { midiToFreq, convertToWav, safeBufferSize } from './helpers';
 import processorNames from './audioWorklet/processorNames';
+import Panner from './panner';
 
 const ac = p5sound.audiocontext;
-
-/**
- *  <p>SoundFile object with a path to a file.</p>
- *
- *  <p>The p5.SoundFile may not be available immediately because
- *  it loads the file information asynchronously.</p>
- *
- *  <p>To do something with the sound as soon as it loads
- *  pass the name of a function as the second parameter.</p>
- *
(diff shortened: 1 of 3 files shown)


src/soundfile.js
```diff
@@ -62,102 +271,1550 @@ const ac = p5sound.audiocontext;
...
+   *    text('tap to download', 10, 20);
+   *  }
+   *
+   *  function canvasPressed() {
+   *    mySound.save('my cool filename');
+   *  }
+   * </code></div>
+   */
+  save(fileName) {
+    p5.prototype.saveSound(this, fileName, 'wav');
```
therewasaguy, line 1750: we should import `saveSound` (perhaps this is handled in an upcoming PR?)

src/soundfile.js
```diff
@@ -62,102 +271,1550 @@ const ac = p5sound.audiocontext;
...
-      if (
-        !(window.File && window.FileReader && window.FileList && window.Blob)
-      ) {
-        // The File API isn't supported in this browser
-        throw 'Unable to load file because the File API is not supported';
+class SoundFile {
+  constructor(paths, onload, onerror, whileLoading) {
+    if (typeof paths !== 'undefined') {
+      if (typeof paths === 'string' || typeof paths[0] === 'string') {
+        var path = p5.prototype._checkFileFormats(paths);
```
therewasaguy, line 278: something for the backlog: we should probably import `checkFileFormats` and remove it from the p5 prototype

therewasaguy: looks like we've started taking care of this in https://github.com/processing/p5.js-sound/pull/540, fantastic!

PR #533, opened 2020-08-20
replacement es5 functions to es6 classes feat p5.Oscillator

endurance21:
@therewasaguy  @kyle1james  tests are smooth 
need a review here ! thanks!

Screenshot from 2020-08-21 02-34-03

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -24,7 +24,6 @@ p5.FFT = FFT;
 import './signal';
 
 import Oscillator, { SinOsc, TriOsc, SawOsc, SqrOsc } from './oscillator';
-console.log(Oscillator);
 p5.Oscillator = Oscillator;
 p5.SinOsc = SinOsc;
 p5.TriOsc = TriOsc;
--- a/src/app.js
+++ b/src/app.js
@@ -22,7 +22,15 @@ import FFT from './fft';
 p5.FFT = FFT;
 
 import './signal';
-import './oscillator';
+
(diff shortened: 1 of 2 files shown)


src/app.js
```diff
@@ -22,7 +22,15 @@ import FFT from './fft';
 p5.FFT = FFT;
 
 import './signal';
-import './oscillator';
+
+import Oscillator, { SinOsc, TriOsc, SawOsc, SqrOsc } from './oscillator';
+console.log(Oscillator);
```
therewasaguy, line 27:
```suggestion

```

src/oscillator.js
```diff
@@ -3,6 +3,41 @@ import Add from 'Tone/signal/Add';
 import Mult from 'Tone/signal/Multiply';
 import Scale from 'Tone/signal/Scale';
 
+// ========================== //
+// SIGNAL MATH FOR MODULATION //
+// ========================== //
+
+// return sigChain(this, scale, thisChain, nextChain, Scale);
+class sigChain {
```
therewasaguy, line 11: I'm not sure if `sigChain` should be a class, since we're not calling it with the `new` keyword, it seems like it can just remain a function.

therewasaguy: calling a method like `add` that calls `sigChain` results in

```
Uncaught TypeError: Cannot call a class as a function
```

endurance21:
@therewasaguy  made some changes  , please have a look!

PR #540, opened 2020-08-20
decoupled helpers method from p5 in helper.js and coupled them separetly in app.js

endurance21:
@therewasaguy @kyle1james  

please review this PR !
okay !  maybe this would be the last one for the gsoc goals !


endurance21:
@therewasaguy  as discussed , we cannot import helpers method out of the src/ folder because of the build system limitations ! 

i think we can go with this ! ?

PR #541, opened 2020-08-24
modernized the  TESTING architecture of the codebase { GSOC} 

endurance21:
## MORE CLEANER AND INTUITIVE TESTING ARCHITECTURE 

* getting rid of requirejs .finally ! 
* used es6 imports 
* used npm to download "chai" and "mocha" and removed the cached version in test/testDeps/chai.js , test/testDeps/mocha.js
thus reducing the `size` of the codebase ! 

* can be further linked to karma.js for headless testing ! { future goals} 
* can be further used for integrating in CI environment !


endurance21:
@therewasaguy @kyle1james  
this one for the testing architecture ! 

for now i have removed sinon.js  ,from test/testDeps/sinon.js ,  i would like to hear some suggestions about its requirements for unit testing now and in future!

test/tests/p5.Amplitude.js
```diff
@@ -1,6 +1,5 @@
 'use strict';
```
therewasaguy, line 1: can we can remove these lines now, too?

test/tests/p5.Amplitude.js
```diff
@@ -64,4 +63,3 @@ define(['chai'], function (chai) {
       }, 10);
     });
   });
-});
```
therewasaguy, line 67: keeping the indentation as-is is very helpful to review, but I think we'll want to format it before we merge.

We can do this with [prettier](https://prettier.io/)

`prettier test/*/js test/tests/*.js --write`

(this could probably be added to our lint task)

test/index.html
```diff
@@ -8,14 +8,18 @@
...
+	
 </head>
 <body>
 	<div id="mocha"></div>
 	<script>
-		mocha.setup('bdd')
+	  	mocha.setup('bdd')
+
 	</script>
+	<script src="./test.js" type="module"></script>
```
therewasaguy, line 23: ✨

therewasaguy:
alright, I think the issue with Firefox is actually an issue with how we've structured some of the tests. For example, I've commented out all of the `import` statements in `test.js` except for `tests/p5.Master.js` and that test runs just fine. Meanwhile `tests/p5.SoundFile.js` has a problem. The issue is that we're creating a p5.Amplitude (which needs an AudioWorklet module) before the p5 init method has completed. We can fix this by creating the amplitude as part of the test lifecyle, wrapping this

https://github.com/processing/p5.js-sound/blob/91d8b34f65e41e29f12cc3e7de17fd12844b717c/test/tests/p5.SoundFile.js#L10

in a `before` method

```
  var a;

  before(() => {
    a = new p5.Amplitude();
  });
```

endurance21:
@therewasaguy  added sinon.js for as it was used inside test/tests/p5.soundRecorder.js

alos removed
``use strict`` keyword

test/tests/p5.Oscillator.js
```diff
@@ -1,87 +1,85 @@
...
+var expect = chai.expect;
 
-  var expect = chai.expect;
+describe('p5.Oscillator', function () {
+  this.timeout(1000);
 
-  describe('p5.Oscillator', function () {
-    this.timeout(1000);
+  var osc = new p5.Oscillator();
+  var amp = new p5.Amplitude();
```
therewasaguy, line 7: I think these need to be instantiated inside of a 'before' lifecycle method (as described in my previous review)

test/tests/p5.SoundFile.js
```diff
@@ -1,142 +1,140 @@
...
+var expect = chai.expect;
 
-  var expect = chai.expect;
+describe('p5.SoundFile', function () {
+  this.timeout(1000);
 
-  describe('p5.SoundFile', function () {
-    this.timeout(1000);
+  var sf;
+  var a = new p5.Amplitude();
```
therewasaguy, line 7: Needs to be wrapped in a `before` method

therewasaguy: (Make sure the tests pass in Firefox, it’s not running for me as-is in Firefox because of this)

endurance21:
@therewasaguy 

this audioWorklet is not loading at the correct time of its access by p5.oscilllator and other processors .
and causing the problem .and mocha hooks is not enough to fix them i see.
This is the problem !
audioWorklet

endurance21:
@therewasaguy  i have solved the problem , 

actually, there was no way we could be able to know when all AudioWorkletProcessor will load finally so I have attached a new 
promise to know  when will all audioWorklet processors will be laoded finally and then used that promise to load the test ! 

the tests are running fine in firefox too ! 

and issue seems to be resolved ! 
have a look then we can move forward 
!

endurance21:
@therewasaguy  hey! it been long seeing  this.
Lets have a good warm up start and we can move ahead for smth good coming year!

montoyamoraga (commented):
hi @endurance21 all of this looks great, thanks!
i'm not familiar with the changes you are doing,
and i don't know how to test it,  so i will let someone else merge it :)

oshoham:
> @therewasaguy i have solved the problem ,
> 
> actually, there was no way we could be able to know when all AudioWorkletProcessor will load finally so I have attached a new
> promise to know when will all audioWorklet processors will be laoded finally and then used that promise to load the test !
> 
> the tests are running fine in firefox too !
> 
> and issue seems to be resolved !
> have a look then we can move forward
> !

Hey @endurance21, the intention of the `AudioWorkletProcessor` loading code that I wrote was that all processor modules would be available when p5's `setup` function runs.

At the time, there was no way to ask p5 to wait for an arbitrary promise before running `setup`, but I was able to use p5's internal preloading hook to load the modules, which is guaranteed to finish before `setup`.

Is it possible for you to just use the p5 `setup` function in your tests instead of adding a new promise that isn't used anywhere else in the codebase?

Alternatively, if p5 has better support for promises these days, maybe you could replace the preloading code entirely?

src/audioWorklet/index.js
```diff
@@ -6,7 +6,10 @@ const moduleSources = [
 ];
 const ac = p5sound.audiocontext;
 
-let initializedAudioWorklets = false;
+let ResolveOnWorkletLoaded;
```
oshoham, line 9: I don't think you need these changes unless you're not able to use the p5 `setup` function in your tests.

src/audioWorklet/index.js
```diff
@@ -28,7 +31,7 @@ p5.prototype.registerMethod('init', function () {
   // use p5's preload system to load necessary AudioWorklet modules before setup()
   this._incrementPreload();
   const onWorkletModulesLoad = function () {
-    initializedAudioWorklets = true;
```
oshoham, line 31: If you're going to remove the `initializedAudioWorklets` variable, can you make sure that the
```
if (initializedAudioWorklets) return;
```
statement on line 25 isn't necessary and if so, remove it?

endurance21:
@oshoham  thanks a ton for the review,  I need to give it a shot by using preload hook and see if that helps.

endurance21:
@oshoham  it worked that way! thanks again!

endurance21:
@therewasaguy  i have tested on firefox and chrome too, tests are running smooth now!

test/tests.js
```diff
@@ -0,0 +1,15 @@
+import('./tests/p5.Master.js');
```
endurance21, line 1: it simple and clean , now we have to write tests in `tests` folder and then import it here!

test/tests/p5.Amplitude.js
endurance21: i think we can have a separate lint task to collectively for all ?

test/tests/p5.SoundFile.js
therewasaguy: also no longer relevant

therewasaguy (approved):
awesome work @endurance21 !

test/index.html
```diff
@@ -1,21 +1,16 @@
...
-</body>
+  <head>
+    <meta charset="utf-8" />
+    <title>TESTS</title>
+    <script src="../lib/p5.js"></script>
+    <script src="../lib/p5.sound.js"></script>
+    <link rel="stylesheet" href="../node_modules/mocha/mocha.css" />
+    <script src="../node_modules/mocha/mocha.js"></script>
+    <script src="../node_modules/chai/chai.js"></script>
+    <script src="../node_modules/sinon/pkg/sinon.js"></script>
```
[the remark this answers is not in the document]
endurance21: earlier those files are stored with fixed version in codebase , now we use npm install and it downloads from npm for every  codebase setup!
that way we just reduced file size of our codebase!

PR #542, opened 2020-08-24
Default to 0 if arraySequence[0] is undefined

MrRoboman:
An error is logged to the console when calling `p5.SoundFile.play()`
Screen Shot 2020-08-23 at 9 24 08 PM

It happens because the SoundFileProcessor sends `[undefined]` to the `ringBuffer` some of the time. 
https://github.com/processing/p5.js-sound/blob/master/src/audioWorklet/soundFileProcessor.js#L18

Causing the `AudioWorkletNode` `onmessage` callback to not fire, which results in `p5.SoundFile.currentTime()` always returning 0.
https://github.com/processing/p5.js-sound/blob/master/src/soundfile.js#L1353-L1364

This change removes the console error and make `currentTime()` behave correctly.

### DIFF
--- a/src/audioWorklet/ringBuffer.js
+++ b/src/audioWorklet/ringBuffer.js
@@ -63,7 +63,7 @@ class RingBuffer {
     // match with this buffer obejct.
 
     // Transfer data from the |arraySequence| storage to the internal buffer.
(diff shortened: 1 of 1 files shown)


MrRoboman:
My pleasure 😄

Issue #529, opened 2020-08-18
p5.score doesn't work


MikhaelOP9:
#### Most appropriate sub-area of p5.js?

- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] Dom
- [ ] Events
- [ ] Image
- [ ] IO
- [ ] Math
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- X Sound

#### Details about the bug:

- p5.js version: /** [p5.sound]  Version: 0.3.12 - 2020-06-26 */ 
- Web browser and version: Firefox 79
- Operating System: Windows
- Steps to reproduce this: the p5.score (sound) doesn't function

therewasaguy:
this should be resolved with #543

PR #543, opened 2020-08-27
fixed issue with p5.Score that prevented 'parts' from being passed as…

aceslowman:
This is related to an issue here: https://github.com/processing/p5.js-sound/issues/529

It doesn't look like p5.Score is currently functional because it fails to fill the 'parts' array with the function arguments, due to a conditional that doesn't seem to be necessary. 

With this said, is there a way to provide a better error message to provide when p5.Score is missing phrases? Currently when p5.Score is provided no arguments, it returns `Cannot read property 'start' of undefined`. I imagine just checking if the function call has any arguments would be enough, but not sure if there is a case where that needs to be allowed.

Also, I would be happy to work on converting this to es6 and writing out some tests, if that would be helpful for this class.


endurance21:
@aceslowman any contribution is most welcome, you can start writing tests after we merge https://github.com/processing/p5.js-sound/pull/541
so that you would able to write test in the new module format, hope patience would  be fruitful soon.

therewasaguy (approved):
thank you @aceslowman ! I just merged in the upstream changes. Look good to you, @endurance21 ?

endurance21:
Looks good to me too.

PR #547, opened 2020-09-19
fix SoundFile.processPeaks()

keksa:
Fixes https://github.com/processing/p5.js-sound/issues/412

There is a `if (!self.panner) return;` check at https://github.com/processing/p5.js-sound/blob/b291d88d9bcb45905852fabbcbb2d8fff0999c00/src/soundfile.js#L1547 which always fails, because `self` variable is undefined.

Not sure if this is proper way to fix this, but it seems pretty obvious and the callback starts to work. 🙂

### DIFF
--- a/src/soundfile.js
+++ b/src/soundfile.js
@@ -1515,6 +1515,7 @@ class SoundFile {
    *  @return {Array}                  Array of timestamped peaks
    */
   processPeaks(callback, _initThreshold, _minThreshold, _minPeaks) {
+    var self = this;
     var bufLen = this.buffer.length;
     var sampleRate = this.buffer.sampleRate;
     var buffer = this.buffer;


endurance21:
@therewasaguy it seems to be a reasonable and effective  approach!

endurance21 (approved):

therewasaguy (approved):
thank you @keksa ! We're considering deprecating this method, however, in https://github.com/processing/p5.js-sound/pull/584

What do you think?

keksa:
@therewasaguy Hi, as I mentioned in https://github.com/processing/p5.js-sound/pull/548#issuecomment-751517090 I don't use this library anymore, so I really have no opinion one way or the other. 🙂

EthanHermsey:
Hey there!

As I understand processPeaks is no longer a thing. Is there a substitute? We used to be able to generate an array of timestamps of peaks/beats of a soundFile beforehand. It seems like you can only get peaks of the live audio output now with peakDetect, is that correct?

PR #548, opened 2020-09-21
p5.soundFile whileLoadingCallback support for FileReader

keksa:
The callback works fine for XHR requests, but the event was not registered for FileReader even though it's also supported by browsers (https://developer.mozilla.org/en-US/docs/Web/API/FileReader/progress_event).

### DIFF
--- a/src/soundfile.js
+++ b/src/soundfile.js
@@ -486,6 +486,9 @@ class SoundFile {
           onerror(e);
         }
       };
+      reader.addEventListener('progress', function (evt) {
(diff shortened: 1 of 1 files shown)


endurance21:
@keksa  thanks for the pr ! , can you mention some of the use cases , where we can get benifit out of it ?

keksa:
@endurance21 If I remember correctly, I used it to display a progress bar while loading a file from file upload input, which does not use XHR. It's been some time though and I abandoned the project in which I was using this library. 🙂

Issue #553, opened 2020-10-25
https://p5js.org/reference/#/p5.AudioIn/getSources 404 error


paulks-software:
404 error on mozilla link
https://p5js.org/reference/#/p5.AudioIn/getSources
and i get , how to list all devices?

This browser does not support MediaStreamTrack.getSources()
(index):86 Uncaught TypeError: Cannot read property 'getSources' of undefined
    at e.div.setup ((index):86)
    at e.<anonymous> (p5.min.js:4)
    at e.<anonymous> (p5.min.js:4)
div.setup @ (index):86
(anonymous) @ p5.min.js:4
(anonymous) @ p5.min.js:4
load (async)
e @ p5.min.js:5
(anonymous) @ (index):102

#### Most appropriate sub-area of p5.js?

- [x ] Accessibility (Web Accessibility)
- [ ] Build tools and processes
- [ ] Color
- [ ] Core/Environment/Rendering
- [ ] Data
- [ ] DOM
- [ ] Events
- [ ] Friendly error system
- [ ] Image
- [x ] IO (Input/Output)
- [ ] Localization
- [ ] Math
- [ ] Unit Testing
- [ ] Typography
- [ ] Utilities
- [ ] WebGL
- [x] Other (specify if possible)

#### Details about the bug:

- p5.js version: 
- Web browser and version: 
- Operating System: 
- Steps to reproduce this:

PaulKsi:
Close it, sorry, found the issue

PaulKsi:
But you can update the website ofc

montoyamoraga:
closed per recommendation of the last poster

PR #566, opened 2020-12-21
fixes issue #553

nakul-shahdadpuri:
Regarding #553. The previous link in docs was broken due to white space thus causing a 404 in the [docs](https://p5js.org/reference/#/p5.AudioIn/getSources). :)

### DIFF
--- a/src/audioin.js
+++ b/src/audioin.js
@@ -270,10 +270,10 @@ class AudioIn {
   }
 
   /**
-   * Returns a list of available input sources. This is a wrapper
-   * for <a href="https://developer.mozilla.org/
-   * en-US/docs/Web/API/MediaDevices/enumerateDevices" target="_blank">
-   * MediaDevices.enumerateDevices() - Web APIs | MDN</a>
+   * Returns a list of available input sources. This is a wrapper for
+   *  <a title="MediaDevices.enumerateDevices() - Web APIs | MDN" target="_blank" href=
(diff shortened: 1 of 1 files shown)


endurance21:
@nakul-shahdadpuri  thanks for raising the pr !
 It would be great if you could present before and after effects of the same , it would save some time ! :)

nakul-shahdadpuri:
Hi @endurance21, thanks for the suggestion. I am adding screenshots of the change below. :).

**Before**
Here clicking on the link lead to a 404. As seen from [here](https://p5js.org/reference/#/p5.AudioIn/getSources).
Screenshot from 2020-12-28 13-59-00

**After**
The link does not lead to a 404.
Screenshot from 2020-12-28 14-42-06

endurance21 (approved):
Looks good to me ! 
Some minor changes , commenting below !

endurance21:
@montoyamoraga 
@therewasaguy 
Have a look at it !

endurance21:
@nakul-shahdadpuri  one things here 
There is a line break between 
Link and 
` it returns a promise ` statement .

I don't think it looks good ?

nakul-shahdadpuri:
@endurance21 cool, I ll remove it right away.

endurance21:
@nakul-shahdadpuri  also , i have seen that , there is rearrangement in title and href and rest is as before , it the problem was with href links should be in same line ?

montoyamoraga (approved):
looks great!

endurance21:
## however , I am holding this pr for now ! because I found some PRs in past  addressing same issue !

Issue #571, opened 2021-01-02
excess onend events during jump / stop+play pairs


rbtcollins:
I'm seeing excess onend events delivered during SoundFile.jump() / stop + play pairs.

PR #572, opened 2021-01-02
Bug #571: fix excess onend events

rbtcollins:
- Allow the BufferSourceNode EventListener to deliver onend when stopAll
  is called. _clearOnEnd will call _onended and remove the
  EventListener.

Signed-off-by: Robert Collins <robertc@robertcollins.net>


therewasaguy (approved):
thank you @rbtcollins !!

PR #574, opened 2021-01-06
Optimized pre-commit-hook action.

frappelatte28:
- Added template in order to run grunt lint only if there are staged changes.
 - Added grunt lint-nofix task to run linter in no-fix mode.
 - Fixed some linting changes too.

I have added the template @endurance21, which is working just like we wanted it to work.

Kindly review the changes @endurance21  @therewasaguy @montoyamoraga @kyle1james  
And just let me know if you need any further changes.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -28,6 +28,13 @@ module.exports = function(grunt) {
           fix: true
         },
         src: ['src/**/*.js', 'test/tests/**/*.js']
+      },
+      sourceNofix: {
+        options: {
+          configFile: './.eslintrc',
+          fix: false
+        },
+        src: ['src/**/*.js', 'test/tests/**/*.js']
       }
     },
     webpack: {
@@ -59,7 +66,10 @@ module.exports = function(grunt) {
     },
     githooks: {
       all: {
-      'pre-commit':'lint' //runs linting test  before every git commit 
+        options:{
+          template:"templates/pre-commit-hook.js"
+        },
+      'pre-commit':'lint-nofix' //runs linting test  before every git commit 
       }
       }
   });
@@ -73,6 +83,7 @@ module.exports = function(grunt) {
   grunt.loadNpmTasks('grunt-githooks');
 
   grunt.registerTask('lint', ['eslint:source']);
+  grunt.registerTask('lint-nofix', ['eslint:sourceNofix']);
   grunt.registerTask('default', ['webpack:prod', 'decomment']);
   grunt.registerTask('dev', ['eslint','connect','webpack:dev', 'decomment']);
   grunt.registerTask('serve', 'connect:server:keepalive');
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -69,7 +69,7 @@ module.exports = function(grunt) {
         options:{
           template:"templates/pre-commit-hook.js"
         },
-      'pre-commit':'lint-nofix' //runs linting test  before every git commit 
+      'pre-commit':'lint-nofix' //runs elint in -nofix mode  before every git commit 
       }
       }
   });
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -28,6 +28,13 @@ module.exports = function(grunt) {
           fix: true
         },
         src: ['src/**/*.js', 'test/tests/**/*.js']
+      },
```
endurance21, line 31: looks good approach. 
btw
There are some common config fields in both `source` and `sourceNofix`, I wonder if we don't have to write them twice?

Gruntfile.js
```diff
@@ -59,7 +66,10 @@ module.exports = function(grunt) {
     },
     githooks: {
       all: {
-      'pre-commit':'lint' //runs linting test  before every git commit 
+        options:{
+          template:"templates/pre-commit-hook.js"
+        },
+      'pre-commit':'lint-nofix' //runs linting test  before every git commit
```
endurance21, line 72: update the comment to 
` runs linter in no fix mode `

templates/pre-commit-hook.js
```diff
@@ -0,0 +1,32 @@
+ // hooks/pre-commit.js
```
endurance21, line 1: oh, that's awesome, seeing this approach!

endurance21 (changes requested):

endurance21:
@frappelatte28  as I can see, files like
-src/ap.js
-src/panner.js
-src/helpers.js

have linting fixes, I do believe that those are necessary changes but we should raise these commits in another PR!
if you are willing to take the shots, go ahead or otherwise tell me if I can be a helping hand.

endurance21:
@frappelatte28  as #575 has been done you can now remove those changes from this PR, then we can see the next thing!
for that, you can either
1. rebase your branch with the current master branch of the repo.
2. or you can just manually reset changes in those files and make a commit for the same in this branch.

frappelatte28:
@endurance21 i have made the changes, waiting for the review!

Gruntfile.js
endurance21: @montoyamoraga  @therewasaguy  do you think we can do smth better here ?

endurance21 (changes requested):

montoyamoraga:
hi, i am sory i cannot help with Gruntfile.js, i am not familiar with how it works

templates/pre-commit-hook.js
```diff
@@ -0,0 +1,32 @@
...
+
+ const exec = require('child_process').exec;
+ // Executes shell commands synchronously
+ const sh = require('child_process').execSync;
+ 
+ exec('git diff --cached --quiet', function (err, stdout, stderr) {
+ 
+  // only run if there are staged changes
+  // i.e. what you would be committing if you ran "git commit" without "-a" option.
+  if (err) {
```
endurance21, line 11: @frappelatte28  i am still wondering why `err` has to be there , should not it be `!err`  ?

endurance21 (changes requested):

templates/pre-commit-hook.js
frappelatte28: yes it is interesting to note that , we are using `--quiet` flag here that reverses the error code to `1` if there are stages changes and `0` if not. 
we can remove `--quiet` and use things as normal.you can look at the reference https://stackoverflow.com/a/28296286

endurance21: Ohh great ! I think that was the concern and it's to get it on board.

endurance21 (approved):

endurance21:
@frappelatte28  great job ! Thanks !

PR #576, opened 2021-02-11
removing p5.signal from codebase (inital)

endurance21:
This is the inital Pr for removing p5.signal from codebase.
Any help on tesing or debugging would really be appreciated.


src/pulse.js
```diff
@@ -1,3 +1,4 @@
+import Signal from 'Tone/signal/Signal';
```
therewasaguy, line 1:
```suggestion
import Signal from 'Tone/signal/Signal';
import Multiply from 'Tone/signal/Multiply';

```

therewasaguy (commented):
thanks for doing on this!

therewasaguy (approved):

endurance21:
I think we are good to go now !
merging this :)

Issue #577, opened 2021-02-12
remove dependency of p5.dom.js from examples 


endurance21:
Many examples are referencing p5.dom.js in the codebase. As p5.dom.js has been merged into p5.js itself. Those references must be removed now.

montoyamoraga:
hi @endurance21 the p5.dom.js was deprecated, and all its functionalities are now in p5.js, so we shouldn't have any p5.dom.js file or dependency.

 my guess is that if we delete any link to p5.dom.js in the examples, they should work now :)

endurance21:
yes , i just saw the same.
Thanks for confirming it

endurance21:
i am modifying this issue by then

satyasaibhushan:
Hey @endurance21!! I'd like to work on

satyasaibhushan:
Hey @endurance21 !!
I'd like to work on this issue.

satyasaibhushan:
So, 
All I have to do is remove the links to p5.dom.js library in all html files in examples. Right?

montoyamoraga:
hi @satyasaibhushan thanks so much for your help! :)

these are my suggested steps:

* make a list of the examples that are trying to import p5.dom.js
* go through them and see if they are broken / how
* remove the p5.dom.js reference
* i'm not sure how the examples are linked to p5.js, maybe check other examples? maybe all examples have a link to p5.js, maybe the link is there but it is pointing to an older version of p5.js and that needs to be updated to the latest one?
* make sure the example is working
* create pull requests, i would make first one PR with one example, to see if the code compiles, the website works, and then small batches of them, because a PR with hundreds of files it's harder to review :)


feel free to ask here any doubt ~

satyasaibhushan:
Hey @montoyamoraga !! thank you for your quick response.
* I think that all the examples are importing the p5.dom.js library through html itself
* I modified first 5 examples (starting from amplitude_analysis ) and they had a 404 error which is fixed by removing the import script statement and worked well with no errors
* So, I'm creating a PR 
###### please correct me if I am doing something wrong

satyasaibhushan:
So, May I do all the examples similarly (but like 20 examples at once) ?

montoyamoraga:
hi @satyasaibhushan i merged your PR, it was great, now let's check on the website if everything is OK, and if so, i say you can do even more examples at once, maybe all of them? it's only one line that you are deleting and i will take care of reviewing the PR :)

[...]

PR #579, opened 2021-02-12
removes depencies of dom-library for first 5 examples #577

satyasaibhushan:
removed import script statement in html form first five examples


montoyamoraga (approved):
looks great!

PR #583, opened 2021-02-24
release : version 1.0.0

endurance21:
we are following semver 2.0 for versioning of the library. 

As we have a deprecated API (p5.Signal) , i have incremented MAJOR from 0 to 1 in the semver format (  MAJOR.MINOR.PATCH )
hence the final version should be  `1.0.0`.
@montoyamoraga @outofambit any suggestions ?


package.json
```diff
@@ -4,7 +4,7 @@
     "type": "git",
     "url": "https://github.com/processing/p5.js-sound.git"
   },
-  "version": "0.3.12",
+  "version": "1.0.0",
```
endurance21, line 7: any suggestion on this semver ?

therewasaguy, line 7: I'm not sure if this feels like a 1.0.0 release, but maybe it's time! Especially if we are going to follow the rules of semver with this non-backwards compatible change (though we haven't been great at following semver for previous releases!).

therewasaguy (approved):

package.json
endurance21: Waving the green flag then ✨.
Off we go 😄❤️

PR #594, opened 2021-03-20
Adding test for helper function

iRohitSingh:
@therewasaguy can you please! review this. I also tried to add the test for the _checkFileFormats but I am not able to debug it. Is there any way to put the breakpoint in test or console.log?

### DIFF
--- a/test/tests/p5.Helpers.js
+++ b/test/tests/p5.Helpers.js
@@ -7,8 +7,8 @@ describe('Testing helpers function', function () {
   });
 
   it('p5.prototype.midiToFreq helper function', function () {
-    const freq = p5.prototype.midiToFreq(720);
-    expect(freq).to.equal(9426054387186970000);
+    const freq = p5.prototype.midiToFreq(100);
+    expect(freq).to.equal(2637.0204553029594);
   });
 
   it('p5.prototype.noteToFreq helper function', function () {
--- a/test/tests/p5.Helpers.js
+++ b/test/tests/p5.Helpers.js
@@ -1,16 +1,35 @@
-var expect = chai.expect;
+const expect = chai.expect;
 
 describe('Testing helpers function', function () {
   it('p5.prototype.freqToMidi helper function', function () {
-    let Midi = p5.prototype.freqToMidi(880);
-    expect(Midi).to.equal(81);
+    const midi = p5.prototype.freqToMidi(880);
+    expect(midi).to.equal(81);
   });
+
   it('p5.prototype.midiToFreq  helper function', function () {
-    let Freq = p5.prototype.midiToFreq(720);
-    expect(Freq).to.equal(9426054387186970000);
+    const freq = p5.prototype.midiToFreq(720);
+    expect(freq).to.equal(9426054387186970000);
   });
+
   it('p5.prototype.noteToFreq  helper function', function () {
-    let Freq = p5.prototype.noteToFreq('C4');
-    expect(Freq).to.equal(261.6255653005986);
+    const freq = p5.prototype.noteToFreq('C4');
+    expect(freq).to.equal(261.6255653005986);
+  });
+
+  it('p5.prototype.soundFormats  helper function', function () {
+    // setting file format so that if we don't provide extension
+    // our file will be loaded because _checkFileFormats add it for us.
+
+    p5.prototype.soundFormats('mp3');
+    const file = p5.prototype._checkFileFormats('a');
+    expect(file).to.be.equal('a.mp3');
+
+    // if we don't provide a valid sound format then soundFormats wil throw
+    //error
+    try {
+      p5.prototype.soundFormats('ext');
+    } catch (err) {
+      expect(err).to.be.equal('ext is not a valid sound format!');
+    }
   });
 });
--- a/test/tests/p5.Helpers.js
+++ b/test/tests/p5.Helpers.js
@@ -6,17 +6,17 @@ describe('Testing helpers function', function () {
     expect(midi).to.equal(81);
   });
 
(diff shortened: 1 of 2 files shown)


iRohitSingh:
@therewasaguy can you please review this, I just wanted to know if I am on right track or not. sorry! to disturb you.

test/tests/p5.Helpers.js
```diff
@@ -0,0 +1,16 @@
+var expect = chai.expect;
+
+describe('Testing helpers function', function () {
+  it('p5.prototype.freqToMidi helper function', function () {
+    let Midi = p5.prototype.freqToMidi(880);
```
therewasaguy, line 5:
```suggestion
    const midi = p5.prototype.freqToMidi(880);
```

test/tests/p5.Helpers.js
```diff
@@ -0,0 +1,16 @@
...
+    expect(Midi).to.equal(81);
+  });
+  it('p5.prototype.midiToFreq  helper function', function () {
+    let Freq = p5.prototype.midiToFreq(720);
```
therewasaguy, line 9:
```suggestion
    const freq = p5.prototype.midiToFreq(720);
```

therewasaguy (changes requested):
really helpful to have these tests, thank you!

Looks like this would work but the convention is to use lower case for variable names, and to use `const` instead of `let` if the variable won't be re-assigned within its scope.

therewasaguy:
> I also tried to add the test for the _checkFileFormats but I am not able to debug it. Is there any way to put the breakpoint in test or console.log?

Great question!

When running the tests (by opening `http://localhost:8000/test` in one or multiple browsers to ensure compatibility across browsers) you should be able to use `console.log` statements.

And you can also set breakpoints in each browser. For example, [here's how to set a breakpoint in Chrome](https://developer.chrome.com/docs/devtools/javascript/breakpoints/
). After opening the JS console, I like to use the shortcut command-shift-p to search for the name of the file where I want to set the breakpoint. Even though the code runs on the compiled `lib/p5.sound.js` file, we can set breakpoints on files from the `/src` directory thanks to the `p5.sound.js.map` file that maps back to the source files.

iRohitSingh:
@therewasaguy thankyou! started amending after your suggestion.

test/tests/p5.Helpers.js
iRohitSingh: Fixed

test/tests/p5.Helpers.js
iRohitSingh: Fixed

iRohitSingh:
@therewasaguy Fixed everything :). I also added a test for the sound formats. is anything left in the helper function because I think other functions are not trivial to test?
waiting for your reply :)

test/tests/p5.Helpers.js
```diff
@@ -0,0 +1,35 @@
+const expect = chai.expect;
+
+describe('Testing helpers function', function () {
+  it('p5.prototype.freqToMidi helper function', function () {
+    const midi = p5.prototype.freqToMidi(880);
+    expect(midi).to.equal(81);
+  });
+
+  it('p5.prototype.midiToFreq helper function', function () {
+    const freq = p5.prototype.midiToFreq(720);
```
therewasaguy, line 10: 720 is outside of the typical MIDI (and frequency) range, which goes from 0 to 127, might be best to use a number within that range?

test/tests/p5.Helpers.js
```diff
@@ -0,0 +1,35 @@
...
+    expect(freq).to.equal(261.6255653005986);
+  });
+
+  it('p5.prototype.soundFormats helper function', function () {
+    // setting file format so that if we don't provide extension
+    // our file will be loaded because _checkFileFormats add it for us.
+
+    p5.prototype.soundFormats('mp3');
+    const file = p5.prototype._checkFileFormats('a');
+    expect(file).to.be.equal('a.mp3');
```
therewasaguy, line 25: This works, but ideally our test won't depend on a method with an underscore prefix (which, according to convention, is private and so could change or be deleted without us needing to update the API)

therewasaguy (commented):
nice! Just a couple suggestions

test/tests/p5.Helpers.js
therewasaguy: can't think of a good way to do this off the top of my head...

test/tests/p5.Helpers.js
iRohitSingh: ok

test/tests/p5.Helpers.js
iRohitSingh: ok! I will try to come with some other function to test soundformat or I will leave it as it is.

test/tests/p5.Helpers.js
iRohitSingh: Done

iRohitSingh:
@therewasaguy I searched throughout the codebase and didn't find a solution other than the check-file format. If you want I can remove it from the test.
can you please merge this I think it is ready?

test/tests/p5.Helpers.js
iRohitSingh: sorry! I am unable to find any better solution other than this.

therewasaguy (approved):

test/tests/p5.Helpers.js
therewasaguy: yeah, I'm not sure either but thanks for considering this feedback

therewasaguy:
thank you @iRohitSingh !

iRohitSingh:
@therewasaguy thank you. This week I will be inactive because I am creating a GSOC proposal for this project.

PR #597, opened 2021-04-08
corrects drywet method in effect class

satyasaibhushan:
I came across this error while running some tests and thought of sharing a draft PR as I want to raise an issue that solves all the errors in the test.


endurance21:
@satyasaibhushan  do you intend to revive it ?

satyasaibhushan:
actually, I just thought of doing a test PR to showcase the problem in my Gsoc proposal.
Do you want me to wait till the projects announcement or Do it right away?

PR #600, opened 2021-04-14
fixed PolySynth  play() bug 

### LINKED ISSUE #599: polysynth play() only works with params more than 2
p5.PolySynth play() member function only works for params more than 2, but in documentation 4 out of 4 params written optional.
```js
  play(note, velocity, secondsFromNow, susTime = 1) {
    this.noteAttack(note, velocity, secondsFromNow);
    this.noteRelease(note, secondsFromNow + susTime);
  }
```
I would like to work on this

Ajaya1000:
fixes #599 
- Changed secondsFromNow to zero
- changed velocity to 0.1

Though it'll work even if the velocity is not set to 0.1, I think it'll make the code much understandable.

### DIFF
--- a/src/polysynth.js
+++ b/src/polysynth.js
@@ -149,7 +149,7 @@ class PolySynth {
    *  }
    *  </code></div>
    */
-  play(note, velocity, secondsFromNow, susTime = 1) {
+  play(note, velocity = 0.1, secondsFromNow = 0, susTime = 1) {
     this.noteAttack(note, velocity, secondsFromNow);
     this.noteRelease(note, secondsFromNow + susTime);
   }


endurance21:
@Ajaya1000 i think velocity's value is being set to 0.1 https://github.com/processing/p5.js-sound/blob/ac73e1d97c78651de8753df925c5e0b197738c93/src/polysynth.js#L259 here if not provided. ?

Ajaya1000:
@endurance21 yes it'll work without the velocity value. But it'll be more understandable that it is optional if we give value to it in the beginning. I can remove it if you want.

therewasaguy (approved):
lgtm and also, @endurance21 that's a good point that velocity is already being set to 0.1 elsewhere.

It could be worth making that value a constant like

`const DEFAULT_VELOCITY = 0.1`

and then we could reference just that constant in both places (in case we want to change it later, we could change it in one place)

therewasaguy:
let's merge it now, but feel free to add that enhancement later if you want @Ajaya1000 and thanks for the PR!

PR #563, opened 2020-12-17
fixes issue #552

nakul-shahdadpuri:
There was an apparent issue with one of the examples on [p5 js reference](https://p5js.org/reference/#/p5.Amplitude). According to the following [issue](https://github.com/processing/p5.js-sound/issues/552), I have made appropriate changes to the example.
Let me know about any changes I should make. :)

PR #602, opened 2021-04-14
Fixed function name mismatch in reference page of p5.Amplitude

endurance21:
# problem
In  PR #563 the name of function was changed but backward compatibility was not ensured.
https://github.com/processing/p5.js-sound/pull/563/commits/498061ee4d54b8886e6273b30df93e1ee5f10730

# solution
Added correct function name at required place.


endurance21:
@therewasaguy these are some errors that is stopping the new releas.
We should adress them asap.
please have a look :)

therewasaguy (approved):
nice catch!

PR #603, opened 2021-04-14
Fixes in Reference Examples 

endurance21:
# Problem
unwanted variable name .
it should be `useStartAudio` instead of `useraudiocontextstart` in inline example of p5.audioin
# Solution 
corrected the variable name


endurance21:
@therewasaguy  quick fix here too

therewasaguy (approved):

PR #604, opened 2021-04-17
Removing var keyword from test code

iRohitSingh:
@therewasaguy I replaced all the var keywords with const and let. I am attaching a screenshot of the test of the main and my branch. Some tests are failing in main that's why it is also failing in my changes also. I will fix it. 
can you please review this? Waiting for your suggestion :)

Screenshot Master
MasterTESTS

Screenshot my branch

MyTest2


therewasaguy (approved):
thank you @iRohitSingh !

PR #605, opened 2021-04-18
Fixing test in monoSynth.js

iRohitSingh:
@therewasaguy The freq of note 'A2' is 55 but we are checking in test 110.

we can test 110 by
```monoSynth.oscillator.freq(110,0.001);```
and then in expect we can do
``` expect(monoSynth.oscillator.freq().value).to.equal(110);```

but I think 55 is also good
This test fails sometime because it compares the default value to 55 if you run the test multiple times I think it is due to Mocha running the function rightaway and our frequency is not set yet.

screenshot
Screenshot from 2021-04-19 02-05-57
 of the passing test

### DIFF
--- a/test/tests/p5.MonoSynth.js
+++ b/test/tests/p5.MonoSynth.js
@@ -12,7 +12,7 @@ describe('p5.MonoSynth', function () {
 
     // wait for scheduled value to complete
     setTimeout(function () {
-      expect(monoSynth.oscillator.freq().value).to.equal(110);
+      expect(monoSynth.oscillator.freq().value).to.equal(55);
       monoSynth.dispose();
       done();
     }, 1);
--- a/test/tests/p5.MonoSynth.js
+++ b/test/tests/p5.MonoSynth.js
@@ -12,7 +12,7 @@ describe('p5.MonoSynth', function () {
 
     // wait for scheduled value to complete
     setTimeout(function () {
-      expect(monoSynth.oscillator.freq().value).to.equal(55);
+      expect(monoSynth.oscillator.freq().value).to.equal(110);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/tests/p5.MonoSynth.js [hunk 1]
```diff
@@ -12,7 +12,7 @@ describe('p5.MonoSynth', function () {
 
     // wait for scheduled value to complete
     setTimeout(function () {
-      expect(monoSynth.oscillator.freq().value).to.equal(110);
```
therewasaguy, line 15: I think this test is correct, an A2 should have a frequency of 110Hz

https://glassarmonica.com/science/frequency_midi.php

And you've caught a bug with the `noteToFreq` method https://github.com/processing/p5.js-sound/blob/4da9b7bb5bc4c18c9fbf7ac1e7857b2770e60a2e/src/helpers.js#L89

The octaves (i.e. the 2 in "A2") start on the C note, so we should make the change in that method to be 

` var wholeNotes = { C: 24, D: 26, E: 28, F: 29, G: 31, A: 33, B: 35 }; `

therewasaguy (changes requested):

test/tests/p5.MonoSynth.js
iRohitSingh: ok

iRohitSingh: fixed

iRohitSingh:
@therewasaguy fixed and test passes. Ready for merge.

iRohitSingh:
@therewasaguy can you please!! review this.

therewasaguy (approved):
thank you @iRohitSingh !

PR #606, opened 2021-04-21
Fixing drywet method bug and p5.Effect.js test

iRohitSingh:
@therewasaguy We are not passing the `fade` parameter to `drywet` method that's why the effect test is failing but now It is fixed.

You can see the passing test in the screenshot

Screenshot from 2021-04-21 23-32-17


therewasaguy:
thank you @iRohitSingh !

therewasaguy (approved):

PR #610, opened 2021-04-25
rename masterVolume to outputVolume

therewasaguy:
- rename masterVolume to outputVolume
- rename `Master` class to `Main`
- update a bunch of examples, tests and comments to match


src/main.js
```diff
@@ -40,18 +41,18 @@ class Master {
...
-const p5sound = new Master();
+// create a single instance of the p5Sound main output for use within this sketch
+const p5sound = new Main();
 
 /**
  * Returns a number representing the master amplitude (volume) for sound
  * in this sketch.
  *
- * @method getMasterVolume
+ * @method outputVolume
```
endurance21, line 51: Should not it be *getOutputVolume*

src/main.js
```diff
@@ -74,14 +75,14 @@ p5.prototype.getMasterVolume = function () {
  *
  *  <p>If no value is provided, returns a Web Audio API Gain Node</p>
  *
- *  @method  masterVolume
+ *  @method  outputVolume
  *  @param {Number|Object} volume  Volume (amplitude) between 0.0
```
endurance21, line 79: volume is repeated 2 times

endurance21 (changes requested):

src/main.js
therewasaguy: yep, this is intentional because the first one gets parsed as the name, the second as part of the description by [jsdoc](https://jsdoc.app/tags-param.html)
jsdoc name and description

src/main.js
therewasaguy: yes, thank you @endurance21 ! Just updated

PR #611, opened 2021-04-28
Test added to p5.fft

endurance21:
added tests for `setInput` method of p5.fft
added some lint fix to  src/looper.js

### DIFF
--- a/test/tests/p5.FFT.js
+++ b/test/tests/p5.FFT.js
@@ -56,54 +56,60 @@ describe('p5.FFT', function () {
     expect(fft.smooth()).to.equal(0.8);
   });
 
-  describe('can set input correctly', function () {
-    let fft;
-
-    beforeEach(function () {
-      fft = new p5.FFT();
-    });
-
-    afterEach(function () {
-      fft.dispose();
-    });
-
-    let soundFile, oscillator;
-    it('can set Oscillator as input', function (done) {
-      this.timeout(1000);
-      oscillator = new p5.Oscillator();
+  let soundFile, oscillator;
+  it('can set Oscillator as input', function (done) {
+    oscillator = new p5.Oscillator();
+    try {
       fft.setInput(oscillator);
-      oscillator.start();
+    } catch (err) {
+      return expect.fail();
+    }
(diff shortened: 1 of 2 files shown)


endurance21:
@therewasaguy please have a look

test/tests/p5.FFT.js
```diff
@@ -55,4 +55,61 @@ describe('p5.FFT', function () {
...
+      done();
+    }, 500);
+  });
+  it('can set soundFile  as input', function (done) {
+    soundFile = p5.prototype.loadSound('./testAudio/drum.mp3', () => {
+      soundFile.disconnect();
+      try {
+        fft.setInput(soundFile);
+      } catch (err) {
+        return done(err);
```
endurance21, line 85: it catches an error , right now i am unsure of why we cant set sounfile as input , any suggestions @therewasaguy

Abhijay007: @davepagurek any thoughts on this?

davepagurek: Interesting, I can't actually reproduce this error, it seems to just work. Maybe it was a browser bug that has since been fixed?

Abhijay007: I will try to reproduce this error, if it works we can merge this one.

Abhijay007: @davepagurek it was dropping these errors for me : 

image

but when I added: `let fft = new p5.FFT();`  it got resolved, anything else seems fine to me.

davepagurek: Oh does this happen when running tests headless from the other open PR? I'm running `npm run test` and going to http://localhost:8000/test/?grep=FFT, which is what shows no errors.

Where are you adding the `new p5.FFT()` line? It also has that in a `beforeEach`, but maybe something is up with the asynchronous tests 🤔

Abhijay007: Ya I am getting that while running the test headless, I added the new line within the function, at the very beginning just like we did for ` oscillator = new p5.Oscillator();`

PR #625, opened 2021-06-10
🔨Fixes noteToFreq function

satyasaibhushan:
A short quick fix correcting the noteToQuick function.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -89,7 +89,7 @@ function noteToFreq(note) {
   var wholeNotes = { A: 21, B: 23, C: 24, D: 26, E: 28, F: 29, G: 31 };
   var value = wholeNotes[note[0].toUpperCase()];
   var octave = ~~note.slice(-1);
-  value += 12 * (octave - 1);
+  value += value < 24 ? 12 * octave : 12 * (octave - 1);
 
   switch (note[1]) {
     case '#':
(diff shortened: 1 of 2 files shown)


src/helpers.js
```diff
@@ -89,7 +89,7 @@ function noteToFreq(note) {
   var wholeNotes = { A: 21, B: 23, C: 24, D: 26, E: 28, F: 29, G: 31 };
   var value = wholeNotes[note[0].toUpperCase()];
   var octave = ~~note.slice(-1);
-  value += 12 * (octave - 1);
+  value += value < 24 ? 12 * octave : 12 * (octave - 1);
```
therewasaguy, line 92: this is a bit confusing to read, compared to changing the values of the first two notes in the `wholeNotes` dictionary in this other PR https://github.com/processing/p5.js-sound/pull/605

`var wholeNotes = { A: 33, B: 35, C: 24, D: 26, E: 28, F: 29, G: 31 };`

therewasaguy (commented):
thank you for the fix, @satyasaibhushan !

Apologies, but I realized that there was already an open PR to fix this test and we should probably merge that one https://github.com/processing/p5.js-sound/pull/605

therewasaguy:
closing as this is now fixed via #605 and again, apologies to both PR's for the delay and duplicated work! Hopefully good learnings along the way at least :)

src/helpers.js
satyasaibhushan: Yeah I agree :)

PR #626, opened 2021-06-11
(Gsoc'21) (Week-1) Fixes broken tests

satyasaibhushan:
This PR fixes the broken tests in the library.
Many of these fixes are small, like adding a timeout.
I've corrected all the tests which were broken in chrome, will add more commits if any of the tests fail in other browsers in the future.


test/tests/p5.PolySynth.js
```diff
@@ -18,15 +18,21 @@ describe('p5.PolySynth', function () {
...
+    await polySynth.play('A4', 0, 0, noteDuration);
+    expect(
+      await polySynth._voicesInUse.getValueAtTime(noteActiveTime)
+    ).to.equal(3);
+
+    expect(await polySynth._voicesInUse.getValueAtTime(noteDoneTime)).to.equal(
+      0
+    );
+
+    await polySynth.dispose();
```
therewasaguy, line 36: I have the same question about all of these `await` operators - I don't think these methods return a Promise and I haven't seen `await` used on methods that don't return a Promise before so I'm not sure the await operator is helping us here. Let me know if I'm missing something though!

test/tests/p5.SoundRecorder.js
```diff
@@ -53,18 +53,21 @@ describe('p5.SoundRecorder', function () {
...
+
+          const outputChannel = outputSoundFile.buffer.getChannelData(0);
+          expect(outputChannel[0]).to.not.eq(0);
+
+          outputSoundFile.dispose();
+          mic.dispose();
+          p5.prototype.outputVolume(0);
+          done();
+        });
+      }, 100);
```
therewasaguy, line 70: why do we need this setTimeout?

test/tests/p5.SoundRecorder.js
```diff
@@ -106,16 +109,18 @@ describe('p5.SoundRecorder', function () {
...
 
-      outputSoundFile.dispose();
-      p5.prototype.outputVolume(0);
-      done();
-    });
+        outputSoundFile.dispose();
+        p5.prototype.outputVolume(0);
+        done();
+      });
+    }, 10);
```
therewasaguy, line 123: same question as above, it seems there is a reason you've wrapped the `recorder.record` function in a setTimeout, I wouldn't think that'd be necessary but maybe it is!

therewasaguy (commented):
awesome, @satyasaibhushan !

I commented with a few questions.

Also, for future PR's that are focused on tests, I'd try to avoid including the re-compiled library as part of the PR commits.

src/soundRecorder.js
```diff
@@ -88,7 +88,7 @@ class SoundRecorder {
     this._inputChannels = 2;
     this._outputChannels = 2; // stereo output, even if input is mono
 
-    const workletBufferSize = safeBufferSize(1024);
+    const workletBufferSize = (this.bufferSize = safeBufferSize(1024));
```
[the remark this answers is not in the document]
satyasaibhushan: Yes, it is being used in [this](https://github.com/processing/p5.js-sound/blob/1b60962f84a3eb68223edf7e74c446cce40399ed/test/tests/p5.SoundRecorder.js#L78) line which is undefined before adding this statement, thus causing the test to fail.

test/tests/p5.Helpers.js
```diff
@@ -8,12 +8,12 @@ describe('Testing helpers function', function () {
 
   it('p5.prototype.midiToFreq helper function', function () {
     const freq = p5.prototype.midiToFreq(100);
-    expect(freq).to.equal(2637.0204553029594);
+    expect(freq).to.be.approximately(2637.0204553029594, 0.00000000001);
```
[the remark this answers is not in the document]
satyasaibhushan: Yeah that works too. I was just going with the flow and kept it precise. Do you want me to change it? I can do it in the next commit which adds to this PR.

test/tests/p5.EQ.js
```diff
@@ -36,22 +36,30 @@ describe('p5.EQ', function () {
...
+  it("a band's freq value can be changed", function () {
     const eq = new p5.EQ(8);
-    expect(eq.bands[0].freq()).to.equal(100);
     eq.bands[0].freq(200);
-    expect(eq.bands[0].gain()).to.equal(0);
-    expect(eq.bands[0].freq()).to.equal(200);
+    setTimeout(() => {
+      expect(eq.bands[0].gain()).to.equal(0);
+      expect(eq.bands[0].freq()).to.equal(200);
+    }, 100);
```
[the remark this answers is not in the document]
satyasaibhushan: Not a reason to pick 100ms exactly, but setting frequencies in this line and this line are takin some time to set the frequency hence we need some timeout. So 0ms would not work.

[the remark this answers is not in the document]
satyasaibhushan: I don't have an exact idea why it does so, but i am fairly confident [this](https://github.com/processing/p5.js-sound/blob/a74dd7eadba27a1e27c46cc7b523faf1b2867dc5/src/filter.js#L165) line may be a reason

test/tests/p5.SoundRecorder.js
satyasaibhushan: This time, the mic is taking some time to connect to the recorder in [this](https://github.com/processing/p5.js-sound/pull/626/files/1b60962f84a3eb68223edf7e74c446cce40399ed#diff-14453cfad7e58ed0e24d64543f6ff750b682d7965d960ff7c8acf40130f19d46R56) line(line 56). If we don't use the timeout, the `outputChannel` is returning all 0's, failing the test.

test/tests/p5.SoundRecorder.js
satyasaibhushan: this situation is similar to the above one. But the Time out here is set to be 10ms as the `outputChannel` without timeout is being all zeros in the returned Float32Array (size 1024) until 287th position and non-zero values after that. The weird part is that the number 287 is being constant no matter how many times I run the code. No idea why this is happening, but the timeout fixes it

test/tests/p5.PolySynth.js
```diff
@@ -18,15 +18,21 @@ describe('p5.PolySynth', function () {
 
     expect(polySynth._voicesInUse.getValueAtTime(noteTriggerTime)).to.equal(0);
 
-    polySynth.play('A2', 0, 0, noteDuration);
+    await polySynth.play('A2', 0, 0, noteDuration);
```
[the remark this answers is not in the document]
satyasaibhushan: Sorry, this is a mistake on my side.

test/tests/p5.PolySynth.js
satyasaibhushan: Hey @therewasaguy, I agree that using await here has no effect. But the problem is that sometimes(like 1 in 4/5) these tests fail as the code is not synchronous. For example,
Screenshot 2021-06-16 at 5 21 35 PM
A quick fix that I can think of is that we layer every statement here with a setTimeout, which would work but will not look good and is inefficient. Can you suggest any other approach or a workaround?

satyasaibhushan:
Commit 3 focuses on correcting the broken tests in firefox.

test/tests/p5.PolySynth.js
therewasaguy: this is a tough one, because [the Web Audio clock is separate (and more precise) than the JS clock](https://www.html5rocks.com/en/tutorials/audio/scheduling/).

I think using `getValueAtTime` (which is not built into the Web Audio API, but comes from using a [Tone.JS](https://tonejs.github.io/docs/14.7.77/Signal#getValueAtTime) Signal and is based on the Web Audio clock) is a good approach. But it only works for parameters that utilize the Tone.JS Signal.

I wonder if the reason it fails sometimes is that we don't pass a raw AudioContext time to PolySynth.play (or to any methods). Instead, we pass a "time from now" of zero. "now" is calculated within the play method here
https://github.com/processing/p5.js-sound/blob/a74dd7eadba27a1e27c46cc7b523faf1b2867dc5/src/polysynth.js#L254 and I wonder how different it is from the "now" that we use in the test?

therewasaguy: for parameters that don't have `getValueAtTime` I think `setTimeout` is reasonable

test/tests/p5.SoundRecorder.js
therewasaguy: got it. It would sure be useful if that method returned a Promise or had a callback

test/tests/p5.EQ.js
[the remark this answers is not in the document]
therewasaguy: ah, that would definitely add some latency!

test/tests/p5.Helpers.js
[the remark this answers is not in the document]
therewasaguy: up to you!

therewasaguy (approved):
thank you, @satyasaibhushan !

test/tests/p5.PolySynth.js
satyasaibhushan: Yeah, 
while exploring that "now" can be different in different cases, I found out that
the "now" in polysyth.js(noteAttack) and the "now" in the test cases is actually coming out to be same as the function is being called while tests are being executed.
But, the time is being different in between the running of tests sometimes, hence sometimes failing the test.
I am adding a commit to this PR which fixes this bug :)

satyasaibhushan:
Hey @therewasaguy, did you also check the commit-3 i.e, part-3 ?

test/tests/p5.PolySynth.js
therewasaguy: yes @satyasaibhushan , interesting solution there in part-4! :shipit:

therewasaguy:
yes, feel free to merge!

PR #628, opened 2021-06-16
(Gsoc'21) (Week-2) Fixes mic issues in recorder example

### LINKED ISSUE #627: Mic issues with soundRecorder example
Currently if we visit [this](https://p5js.org/reference/#/p5.SoundRecorder) page, we get a prompt asking for the mic permission and it is kept on even if we do not interact with the example which is not necessary. Even if we interact with the example and finish hearing it, the browser will not stop accessing the mic. 

I think it would be good if the mic is accessed only while interacting with the example as the user might be on this page only because of the reference but not example.

satyasaibhushan:
Fixes #627,
Fixed the example in both the examples folder's file and also the inline code.

To make sure that the mic is called only after the user's interaction with the example, I've moved the mic setup code to the first click.
Also disposed the mic after the recording is completed.

##### Also adjusted the comments in the inline code accordingly and verified it by building locally.

### DIFF
--- a/examples/record/sketch.js
+++ b/examples/record/sketch.js
@@ -2,9 +2,10 @@
 // We need p5.AudioIn (mic / sound source), p5.SoundRecorder
 // (records the sound), and a p5.SoundFile (play back).
 
-var mic, recorder, soundFile;
+let mic, recorder, soundFile;
 
-var state = 0; // mousePress will increment from Record, to Stop, to Play
+let state = 0; // mousePress will increment from state = 0 (Record), to 1(Stop), to 2(Play)
+
 
 function setup() {
   createCanvas(400, 400);
@@ -26,10 +27,11 @@ function setup() {
 }
 
 function mousePressed() {
+  userStartAudio();
   // use the '.enabled' boolean to make sure user enabled the mic (otherwise we'd record silence)
   if (state === 0) {
     // users must manually enable their browser microphone for recording to work properly!
-    mic.start(function() {
+    mic.start(function () {
       // Tell recorder to record to a p5.SoundFile which we will use for playback
       recorder.record(soundFile);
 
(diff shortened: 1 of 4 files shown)


examples/record/sketch.js
```diff
@@ -7,17 +7,14 @@ var mic, recorder, soundFile;
 var state = 0; // mousePress will increment from Record, to Stop, to Play
```
endurance21, line 7: Converting vars to let would be a good improvement in code style.
Also the variable name state =0 is not so semantic about what it is doing ,
Maybe a string like 
let recordingStarted = true/false 
Would make sense or smth like that

endurance21 (changes requested):

examples/record/sketch.js
satyasaibhushan: Agree with the vars to let part

But regarding the renaming of state, we need to cover a total of 3 cases. Hence, it cannot be bool.
We can overcome this in two ways
1. we can maintain 2 booleans like isRecordingStarted and isResultPlayed and use them both to get 3 cases
2. We can use something like a string, but we should set the value of the string after each conditional and use it as a condition for next step

I think 2nd case would not be better than the current situation.It would be better if we change it to case 2, which can be debatable
Please share your thoughts/opinions on this :)

endurance21: I think using multiple books would be good instead of using string .
Let's just keep in sleak and intuitive with bools 
:)

guillemontecinos, line 7: I agree with @endurance21 on updating `var` to `let`.
Regarding the `state` variable, since it was originally defined that way I would touch the logic of the example. What I definitely suggest is to make the comment after the definition of `state` more explicit, like:

```js
// mousePress will increment from state = 0 (Record), to state = 1 (Stop), to statue = 2 (Play)
```

satyasaibhushan: Hey I've implemented the 2 bools in a separate commit.
If you like that approach better, you can include while merging or else you can leave it out

satyasaibhushan: Yeah,
I've implemented it in the 2nd commit,
if the 2 boolean approach is not merged and we are left with a state variable, it will be retained in the code

endurance21 (approved):

PR #632, opened 2021-07-11
(Gsoc'21) (Week-3,4) Adding new tests for uncovered files

satyasaibhushan:
Added new unit tests for all the uncovered 16 files.

I left some tasks/ tests to be written as a todo in some files, which I will get back to in a future PR.


test/tests/p5.Gain.js
```diff
@@ -0,0 +1,51 @@
...
+
+  describe('methods', function () {
+    describe('setInput', function () {
+      it('can set Input', function () {
+        let soundFile = p5.prototype.loadSound('./testAudio/drum.mp3');
+        gain.setInput(soundFile);
+      });
+    });
+    describe('connect, disconnect', function () {
+      it('can connect to empty node', function () {
```
therewasaguy, line 31:
```suggestion
      it('connects to p5.soundOut when no arguments are provided', function () {
```

just observing that this is what happens when no arguments are provided

https://github.com/processing/p5.js-sound/blob/17987e714cf5bce6904e31912603a0d640e8d0c9/src/gain.js#L106-L107

test/tests/p5.Gain.js
```diff
@@ -0,0 +1,51 @@
...
+    gain = new p5.Gain();
+  });
+
+  it('can be created', function () {
+    expect(gain.input).to.have.property('gain');
+    expect(gain.output).to.have.property('gain');
+    let audioContext = gain.ac;
+    expect(audioContext).to.have.property('baseLatency').to.be.an('number');
+    expect(audioContext).to.have.property('destination');
+    expect(audioContext).to.have.property('state').to.be.an('string');
```
therewasaguy, line 15: these look like the same `expect`'s that were part of the AudioContext test. I wonder if they could be abstracted to a method wherever we need to verify that there is an AudioContext.

test/tests/p5.Metro.js
```diff
@@ -0,0 +1,85 @@
...
+      metro.stop();
+      expect(ticks).to.not.equal(0);
+      setTimeout(() => {
+        expect(metro.metroTicks).to.equal(ticks);
+        done();
+      }, 100);
+    }, 1000);
+  });
+
+  it('can be started and stopped with dealy', function (done) {
```
therewasaguy, line 39:
```suggestion
  it('can be started and stopped with delay', function (done) {

File: test/tests/p5.Metro.js [hunk 4]
```diff
@@ -0,0 +1,85 @@
...
+    let ticks;
+    metro.setBPM(600);
+    metro.start(0.1);
+    setTimeout(() => {
+      ticks = metro.metroTicks;
+      metro.stop(0.2);
+      expect(ticks).to.not.equal(0);
+      setTimeout(() => {
+        expect(metro.metroTicks).to.not.equal(ticks);
```
therewasaguy, line 48:
```suggestion
        expect(metro.metroTicks).to.be.above(ticks);
```

In an ideal world we would test that the value is equal to an exact value. But since we are using the JS clock's setTimeout, which is not in sync or as precise as the AudioContext time, we can't really know what that value will be, but "greater than" feels more specific than "not equal" which could mean lower or higher.

test/tests/p5.Metro.js
```diff
@@ -0,0 +1,85 @@
+const expect = chai.expect;
+let metro;
+
+describe('p5.Metro', function () {
+  beforeEach(function () {
+    metro = new p5.Metro();
+  });
```
therewasaguy, line 7: it might help with garbage collection to add an `afterEach` to call the "dispose" method which removes references to the Web Audio nodes freeing up those resources for garbage collection.

But if things are running just fine so far, maybe that's not necessary.

test/tests/p5.PeakDetect.js
```diff
@@ -0,0 +1,21 @@
+const expect = chai.expect;
+
+describe('p5.PeakDetect', function () {
+  it('can be initalized without any arguemts', function () {
```
therewasaguy, line 4:
```suggestion
  it('can be initialized without any arguments', function () {

File: test/tests/p5.Looper.js [hunk 7]
```diff
@@ -0,0 +1,345 @@
+const expect = chai.expect;
+
+describe('p5.Looper', function () {
```
therewasaguy, line 3: really appreciate the detailed unit tests for this class, you dug deep here!

test/tests/p5.Noise.js
```diff
@@ -0,0 +1,44 @@
...
+      expect(noise).to.not.have.property('noise');
+      noise.start();
+      expect(noise).to.have.property('noise');
+      expect(noise.noise).to.have.property('buffer');
+      expect(noise.noise.loop).to.be.true;
+      expect(noise.started).to.be.true;
+      noise.stop();
+      expect(noise.started).to.be.false;
+    });
+    //TODO: test noise buffer generator functions
```
therewasaguy, line 42: probably out of scope for the summer, but this could be a fun project

test/tests/p5.Pulse.js
```diff
@@ -0,0 +1,57 @@
...
+    expect(pulse.dcOffset).to.have.property('buffer');
+    expect(pulse.dcOffset).to.have.property('channelCount');
+    expect(pulse.dcGain).to.have.property('gain');
+    expect(pulse.output.gain.value).to.equal(1);
+  });
+  it('can be with arguments', function () {
+    let pulse = new p5.Pulse(220, 0.4);
+    expect(pulse.w).to.equal(0.4);
+    expect(pulse.f).to.equal(220);
+    console.log(pulse);
```
therewasaguy, line 21:
```suggestion

File: test/tests/p5.Pulse.js [hunk 10]
```diff
@@ -0,0 +1,57 @@
...
+      pulse.start(221, 0.1);
+      setTimeout(() => {
+        expect(pulse.oscillator.frequency.value).to.equal(221);
+        expect(pulse.oscillator.type).to.equal('sawtooth');
+        expect(pulse.osc2.oscillator.type).to.equal('sawtooth');
+        done();
+      }, 500);
+      expect(pulse.started).to.be.true;
+      expect(pulse.osc2.started).to.be.true;
+      pulse.stop(221, 0.1);
```
therewasaguy, line 49:
```suggestion
      pulse.stop()
```
it doesn't seem worth testing with delay if the `.started` property that we are testing with below doesn't respect the delay

test/tests/p5.SoundLoop.js
```diff
@@ -0,0 +1,161 @@
...
+      setTimeout(() => {
+        ticks = sloop.iterations;
+        sloop.stop();
+        expect(sloop.isPlaying).to.be.false;
+        expect(ticks).to.not.equal(0);
+        expect(count).to.be.approximately(ticks, 2);
+        done();
+      }, 1000);
+    });
+    it('can be started and stopped with dealy', function (done) {
```
therewasaguy, line 51:
```suggestion
    it('can be started and stopped with delay', function (done) {
```

test/tests/p5.Envelope.js
```diff
@@ -0,0 +1,270 @@
...
+    });
+    it('can trigger an attack at the present moment', function () {
+      let envelope = new p5.Envelope(0.65, 0.75, 0.85, 0.55, 0.95, 0.9);
+      let osc = new p5.SinOsc();
+      let now = p5.prototype.getAudioContext().currentTime;
+      envelope.triggerAttack(osc);
+      expect(envelope.lastAttack).to.not.be.zero;
+      expect(envelope.wasTriggered).to.be.true;
+
+      expect(envelope.control.getValueAtTime(now + 0.65)).to.equal(0.75);
```
therewasaguy, line 134: good idea to utilize the `getValueAtTime` method where possible here!

therewasaguy (approved):
@satyasaibhushan amazing to see how deeply you've gotten to know the library in order to write these tests (and fix a lot of bugs along the way!)

I left some comments with tiny suggestions, and the larger ones are probably outside the scope of GSoC at this point!

test/tests/p5.Gain.js
satyasaibhushan: Yep, looks good

test/tests/p5.Metro.js
satyasaibhushan: I am surprised to find out that metro does not have a dispose function. What should I do?
may be clear out the variable?

test/tests/p5.SoundLoop.js
satyasaibhushan: Sorry for the spelling mistakes :}

test/tests/p5.Metro.js
satyasaibhushan: Agreed

test/tests/p5.Pulse.js
satyasaibhushan: Opps

test/tests/p5.Metro.js
therewasaguy: ah, that sounds like an issue that should be handled in a separate PR, then!

test/tests/p5.SoundLoop.js
therewasaguy: no worries, I'm also prone to typos in strings/comments. It might help to add a spell check plugin for your IDE

therewasaguy (approved):
fantastic!

test/tests/p5.Noise.js
satyasaibhushan: Yeah, I didn't know how to test them.

test/tests/p5.Panner.js
```diff
@@ -0,0 +1,33 @@
...
+  it('can be connected and disconnected', function () {
+    let panner = new p5.Panner(input, output);
+    panner.connect(input);
+    panner.disconnect();
+  });
+  it('can be panned without a delay', function () {
+    let panner = new p5.Panner(input, output);
+    panner.pan(0.4);
+    panner.pan(0.3, 0);
+    //TODO: to check the value of left gain/ right gain (or) the stereoPanner.pan
```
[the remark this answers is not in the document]
satyasaibhushan: Yeah, I tried but the function setValueAtTime is difficult to test as the value is set at that particular time.

test/tests/p5.Gain.js
satyasaibhushan: Yes, actually. We would use that very often. also we can add similar functions for checking gain nodes and etc.

Will add it if I got time or will at least add a issue/TODO regarding this.

PR #636, opened 2021-07-29
(Gsoc'21) (Week-6,7)Adding tests for covered files 

satyasaibhushan:
Added unit tests for all the 18 files which are partially covered 

This PR is built upon the [previous](https://github.com/processing/p5.js-sound/pull/632) PR.

I also fixed some potential bugs in some commits.

I followed a particular style (hierarchy ) for all the tests written in this period, which can be changed according to the suggestions.


satyasaibhushan:
This PR consists of 32 commits ,11 from the previous PR, 2 for merging , 19 for adding tests to covered files.

satyasaibhushan:
Hey @therewasaguy , me and @guillemontecinos agreed that it would not be possible for one person to review all the changes and that it would be better if we split the files to be reviewed among the three reviewers suggested. What do you think?
 
 Also, there were like 6-8 commits in which I fixed some potential minor bugs for which I need your thoughts on.
 So, it would be great if you assign those commits to yourself. These commits can be identified with a hammer symbol (🔨) before them.

satyasaibhushan:
Hey @therewasaguy, Please complete this as soon as possible.
Also, I tried removing the repeated commits (commits with 🔨 ) but could not do that. Either we can ignore those commits as repeated or you could suggest me doing something.

src/amplitude.js
```diff
@@ -184,9 +184,10 @@ class Amplitude {
       } else {
         this.output.connect(unit);
       }
-    } else {
-      this.output.connect(this.panner.connect(p5sound.input));
     }
+    // else {
+    //   this.output.connect(this.panner.connect(p5sound.input));
+    // }
```
therewasaguy, line 190: it's not clear to me why this was commented out. It looks like this is meant to be the default behavior of the `connect` method when no arguments are passed in, and I don't see that in the unit tests so it's not clear to me why this would be commented out (and better to delete it if it is meant to be commented out!)

src/fft.js
```diff
@@ -171,13 +171,12 @@ class FFT {
...
    */
   waveform() {
-    var bins, mode;
+    var mode;
     var normalArray = new Array();
 
     for (var i = 0; i < arguments.length; i++) {
       if (typeof arguments[i] === 'number') {
-        bins = arguments[i];
-        this.analyser.fftSize = bins * 2;
```
therewasaguy, line 180: nice improvement here, since `this.bins` has a setter that also updates `this.analyser.fftSize` 💯 
https://github.com/processing/p5.js-sound/blob/17987e714cf5bce6904e31912603a0d640e8d0c9/src/fft.js#L92-L98

test/tests/main.js
```diff
@@ -1,10 +1,36 @@
...
+    expect(p5.soundOut.fftMeter).to.have.property('context');
+    expect(p5.soundOut.soundArray).to.be.an('array');
+    expect(p5.soundOut.parts).to.be.an('array');
+    expect(p5.soundOut.extensions).to.be.an('array');
+
+    expect(p5.soundOut._silentNode).to.have.property('gain');
+    expect(p5.soundOut._silentNode).to.have.property('context');
+    expect(p5.soundOut._silentNode.gain.value).to.equal(0);
+
+    console.log(p5.soundOut);
```
therewasaguy, line 25:
```suggestion
```

some repos have a linting rule to remove console.logs which I always find helpful to clean up my own 😄

therewasaguy (commented):
This is awesome, @satyasaibhushan!

Just one small comment/question from me. I do not have the bandwidth to go through all of the tests in detail before GSoC ends, but I think that this is clearly a huge improvement, cc @guillemontecinos feel free to approve and merge!

src/amplitude.js
satyasaibhushan: Yeah, I agree that it is the default behaviour. But it's just that there is no panner node for amplitude(may be removed). To remove the error, I just commented it

Do you want me to remove the lines or just post an issue?

guillemontecinos (approved):

src/amplitude.js
therewasaguy: ah, good catch!

This sounds like a separate issue then, "Amplitude connect method with no arguments should connect to main output." 

In general, it's best not to leave commented-out code in the code base.

PR #639, opened 2021-08-09
Bug fixing

satyasaibhushan:
I isolated the commits from my previous 2 PRs(week-3,4 week-6,7) that consists of various bug fixing that I felt are potential bugs.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -88,8 +88,8 @@ p5.BandPass = BandPass;
 import EQ from './eq';
 p5.EQ = EQ;
 
-import listener3D from './listener3d';
-p5.listener3D = listener3D;
+import Listener3D from './listener3d';
+p5.Listener3D = Listener3D;
 
 import Panner3D from './panner3d';
 p5.Panner3D = Panner3D;
--- a/src/looper.js
+++ b/src/looper.js
@@ -387,12 +387,12 @@ class Score {
   constructor() {
     // for all of the arguments
     this.parts = [];
-    this.currentPart = new Array(arguments.length);
+    this.currentPart = 0;
 
     var thisScore = this;
     for (var i in arguments) {
       this.parts[i] = arguments[i];
-      this.parts[i].nextPart = this.parts[i + 1];
+      if (i > 0) this.parts[i - 1].nextPart = this.parts[i];
       this.parts[i].onended = function () {
         thisScore.resetPart(i);
         playNextPart(thisScore);
--- a/src/looper.js
+++ b/src/looper.js
@@ -392,7 +392,9 @@ class Score {
     var thisScore = this;
     for (var i in arguments) {
       this.parts[i] = arguments[i];
-      if (i > 0) this.parts[i - 1].nextPart = this.parts[i];
+      if (i > 0) {
+        this.parts[i - 1].nextPart = this.parts[i];
+      }
       this.parts[i].onended = function () {
         thisScore.resetPart(i);
         playNextPart(thisScore);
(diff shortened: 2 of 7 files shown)


src/reverb.js
```diff
@@ -103,20 +103,7 @@ class Reverb extends Effect {
...
-    if (decayRate) {
-      this._decay = decayRate;
-    }
-    if (reverse) {
-      this._reverse = reverse;
-    }
-    if (rebuild) {
-      this._buildImpulse();
-    }
+    this.set(seconds, decayRate, reverse);
```
therewasaguy, line 106: I'm curious what bug this fixes? It would be helpful to see the corresponding test that helped you identify this as a solution.

In any case, good call to re-use the method rather than duplicate code!

src/app.js
```diff
@@ -88,8 +88,8 @@ p5.BandPass = BandPass;
 import EQ from './eq';
 p5.EQ = EQ;
 
-import listener3D from './listener3d';
-p5.listener3D = listener3D;
```
therewasaguy, line 92: yikes, great catch here!

src/looper.js
```diff
@@ -387,12 +387,12 @@ class Score {
...
     // for all of the arguments
     this.parts = [];
-    this.currentPart = new Array(arguments.length);
+    this.currentPart = 0;
 
     var thisScore = this;
     for (var i in arguments) {
       this.parts[i] = arguments[i];
-      this.parts[i].nextPart = this.parts[i + 1];
+      if (i > 0) this.parts[i - 1].nextPart = this.parts[i];
```
therewasaguy, line 395:
```suggestion
      if (i > 0) {
        this.parts[i - 1].nextPart = this.parts[i];
      }
```

nice solution/bug fix here!

I prefer to use brackets but that's just my personal preference for if statements

src/polysynth.js
```diff
@@ -389,6 +387,7 @@ class PolySynth {
         this.notes[n].dispose();
         delete this.notes[n];
       }
+      this._newest = 0;
```
therewasaguy, line 390: I'm curious what test that you added to uncover this bug (I can take a look at your other PR to find out! But in hindsight / next time, it would've been useful to include the unit tests and bug fixes in the same PR)

therewasaguy (approved):
really good finds and fixes, @satyasaibhushan !

therewasaguy:
Thank you for breaking the PR into a smaller piece that was a bit easier to review.

In hindsight, it would've been useful to see the unit tests that helped you uncover the bugfix as part of the same PR—perhaps one PR per unit would've been a nice way to break it down. Something to consider for future PR's.

In any case, this was a bit less daunting than #636 as I've just returned from a couple weeks away from the computer, but I hope we can also get all of your amazing work reviewed and merged soon!

src/reverb.js
satyasaibhushan: Yeah, Sorry for that. It does not fix any bug. Just rearranged the code

src/polysynth.js
satyasaibhushan: I do not think a particular test uncovered this. While testing the function to include a case that if a note is not provided, it should reset all the notes. It seemed logical to reset the newest value as all the notes are being erased

satyasaibhushan: Actually, I think we should also reset the oldest value also right?

satyasaibhushan:
Hey, thanks for the review.
If you agree with the last 2 commits, you can go ahead and merge this PR.
Also, keep in mind that you should first review #632 before #636, as it is built on top of the previous one.

PR #646, opened 2021-10-07
Add Listener 3d tests

Devesh21700Kumar:
Add tests to check if it `can set upX, forwardY, upZ with a delay`
for the p5.Listener3D instance


Abhijay007 (commented):
@Devesh21700Kumar code LGTM but can you please un-stage the files other than `test/tests/p5.Listener3d.js` as changes of those files are not required?

davepagurek (approved):
Thanks for adding the tests @Devesh21700Kumar! I'm taking out the `lib/` changes because we'll do those once we release to keep PRs easier to digest.

Issue #672, opened 2022-09-08
p5.SoundFile addCue - incorrect documentation, time is not passed to the callback


sialivi:
The documentation states:
> "Time will be passed as the first parameter to the callback function, and param will be the second parameter."

As far as I can tell (I'm new to p5 and javascript) only param is passed to the callback.

PR #683, opened 2022-11-22
documentation issue #672 re: callback params in p5.SoundFile addCue

capogreco:
Changed the documentation to reflect that the third argument is given to the callback as its only parameter (the time can be incorporated into that argument if required).

issue is here: https://github.com/processing/p5.js-sound/issues/672


Abhijay007 (approved):
LGTM

davepagurek (approved):
thanks for the change! This wording is indeed clearer.

PR #698, opened 2022-12-22
fix inconsistency in documentation of effects

KevinGrajeda:
- renamed from "Signal" to "src" in Delay and Filter `process()` method documentation.
- added "This class extends [p5.Effect](https://p5js.org/reference/#/p5.Effect). Methods [amp()](https://p5js.org/reference/#/p5.Effect/amp), [chain()](https://p5js.org/reference/#/p5.Effect/chain), [drywet()](https://p5js.org/reference/#/p5.Effect/drywet), [connect()](https://p5js.org/reference/#/p5.Effect/connect), and [disconnect()](https://p5js.org/reference/#/p5.Effect/disconnect) are available." missing text in reverb and panner3d.
- added missing effects in the effect documentation.

### DIFF
--- a/src/delay.js
+++ b/src/delay.js
@@ -114,7 +114,7 @@ class Delay extends Effect {
    *
    *  @method  process
    *  @for p5.Delay
-   *  @param  {Object} Signal  An object that outputs audio
+   *  @param  {Object} src An object that outputs audio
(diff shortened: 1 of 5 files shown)


davepagurek (approved):
Great work, thanks for adding those extra links!

PR #706, opened 2022-12-29
panValue is not optional

Nitesh639:
Pan accepts two values ​​in SoundFile class, where the timeFromNow is optional but panValue is not optional.


davepagurek (approved):
Make sense, thanks for clarifying this!

Nitesh639:
Thank you.

PR #707, opened 2022-12-29
example reference added

Nitesh639:
Now example reference is added for the pan in the oscillator class.


davepagurek (approved):
Nice!

Nitesh639:
Thank you.

PR #708, opened 2022-12-30
start_stop doc

Nitesh639:
No reference was provided for userStartAudio(). And also the secondsFromNow is an optional parameter in p5.Oscillator.stop() method.
As we can see [here](https://github.com/processing/p5.js-sound/blob/main/src/oscillator.js#L207).


Abhijay007 (approved):
LGTM 👍

davepagurek (approved):
Looks good!

PR #711, opened 2022-12-31
fix compressor.reduction()

### LINKED ISSUE #710: Compressor.reduction() always returns undefined
When creating a Compressor, `reduction()` should return the current reduction value but always returns undefined.

example code:
```JavaScript
function setup() {
  let compressor=new p5.Compressor();
  console.log(compressor.reduction()); //prints: undefined, should print: 0
}
```

KevinGrajeda:
Resolves #710

- `Compressor.reduction()` now returns `this.compressor.reduction` instead of `this.compressor.reduction.value` since reduction doesn't have a value property
- test now uses `reduction()` instead of directly accesing the property with `compressor.compressor.reduction.value`

### DIFF
--- a/src/compressor.js
+++ b/src/compressor.js
@@ -235,7 +235,7 @@ class Compressor extends Effect {
    * @return {Number} Value of the amount of gain reduction that is applied to the signal
    */
   reduction() {
-    return this.compressor.reduction.value;
+    return this.compressor.reduction;
   }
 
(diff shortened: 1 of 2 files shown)


davepagurek (approved):
Just double checked the specs, and agreed, this property is just a number and not an `AudioParam`. Thanks!

PR #719, opened 2023-01-05
add example and documentation to Noise 

KevinGrajeda:
- Added new example to p5.Noise 
- setType() 'type' is now not documented as optional
- Added documentation to `getType()`, `start()`, `stop()` and `dispose()`


davepagurek (approved):
This is a really cool example, great work!

PR #720, opened 2023-01-06
(Chore) : Fixed merge conflicts in "fixing EQ test"

Abhijay007:

When we pass 8 in the EQ constructor it takes some time to initialize the freq but we are checking just after the setting that's why it is comparing with the default value i.e 350. Same issue when we are setting the freq to 200. 

Resolved merge conflicts in PR https://github.com/processing/p5.js-sound/pull/615

117196347-fd230e00-ae03-11eb-96c0-b8b066d1ba23

### DIFF
--- a/test/tests/p5.EQ.js
+++ b/test/tests/p5.EQ.js
@@ -54,13 +54,17 @@ describe('p5.EQ', function () {
           }, 100);
         });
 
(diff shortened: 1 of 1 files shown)


davepagurek (approved):
Thanks, this change makes sense!

Issue #665, opened 2022-05-25
getEnergy() throws error with zero as first argument, but not for second, and not for negative numbers


awarua:
The `getEnergy()` function throws an 'invalid input for getEnergy()' error if given zero as a first argument. I don't really know anything about the technicalities of how the FFT function works, but I guess since an error is being thrown that this is expected behaviour?
* `let e = fft.getEnergy(0, 140);`

However, passing zero as the second argument and passing negative numbers **do not** throw an error. 

Passing negative numbers seems to return `NaN`. 
* `let e = fft.getEnergy(-20, 140);`

Passing zero as second argument returns `undefined`
* `let e = fft.getEnergy(-20, 0);`

It would be good to make the error throwing behaviour consistent and also to update the reference page to tell users that they should not pass zero as an argument to `getEnergy()`. I came across this because users in a class I teach tried it with `getEnergy(0, 140)` as arguments, which I think is a reasonable expectation.

PR #722, opened 2023-01-12
fix FFT.getEnergy parameter validation

KevinGrajeda:
Addresses #665 (I am not closing the issue because I am not sure what should be done with negative numbers)

- `getEnergy()` now accepts 0 as a parameter
- Removed extra `else` statements
- Removed redundant `if`

### DIFF
--- a/src/fft.js
+++ b/src/fft.js
@@ -350,9 +350,6 @@ class FFT {
       var index = Math.round((frequency1 / nyquist) * this.freqDomain.length);
       return this.freqDomain[index];
     }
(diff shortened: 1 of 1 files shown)


KevinGrajeda:
I don't know if throwing an error when using negative numbers would be a good idea.

davepagurek (approved):
Thanks for making these changes!

0 is a valid frequency to get energy of (it's a DC offset on the signal), but we won't ever have negative frequencies, so I think it makes sense to throw an error in that case (and maybe mention in the docs that the frequency input must be >= 0.)

PR #730, opened 2023-01-17
add negative number validation to `getEnergy()`

### LINKED ISSUE #665: getEnergy() throws error with zero as first argument, but not for second, and not for negative numbers

KevinGrajeda:
Closes #665 

- Added validation for negative frequency values
- added 'frequency must be >= 0' to `getEnergy()` docs
- removed duplicate 'Energy' in docs

### DIFF
--- a/src/fft.js
+++ b/src/fft.js
@@ -298,7 +298,7 @@ class FFT {
    *  <a href="https://en.wikipedia.org/wiki/Audio_frequency" target="_blank">
    *  frequency</a>, or the average amount of energy between two
    *  frequencies. Accepts Number(s) corresponding
-   *  to frequency (in Hz), or a "string" corresponding to predefined
+   *  to frequency (in Hz) (frequency must be >= 0), or a "string" corresponding to predefined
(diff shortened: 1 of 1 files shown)


davepagurek (approved):
Looks great, thanks!

PR #738, opened 2023-01-26
add `_onNewInput()` hook 

### LINKED ISSUE #737: `_onNewInput()` hook
As discussed in #723, we want to add an  `_onNewInput()` hook to let nodes know they've been connected to, but to do this, it is necessary to change every `connect()` method.
At the end of each `connect()` we have to add:
```javascript
connect(unit) {
  //  ...
 
  if (unit._onNewInput) {
    unit._onNewInput(this);
  }
}
```
I think the classes that need this change are:

1. `AudioIn.js`
2. `AudioVoice.js`
3. `Effect.js`
4. `Envelope.js`
5. `Gain.js`
6. `MonoSynth.js`
7. `Oscillator.js`
8. `Panner.js`
9. `PolySynth.js`
10. `SoundFile.js`

from p5.js, the `MediaElement.connect()` method can be used with p5-sound objects, so in the future i think we would like to have this change in that method.

`Amplitude.js` also has a `connect()` method but I don't know what its use is so I don't know if it should be changed.
What should be done with these last two classes? and are those all the classes that need the change?

KevinGrajeda:
Closes #737 

- Classes with `connect(unit)` method now call `unit._onNewInput()` .
- Added `FFT._onNewInput()` so that FFT.js stops listening to all sound in the sketch when something is connected to it.
- Added tests for `._onNewInput()`.

### DIFF
--- a/src/amplitude.js
+++ b/src/amplitude.js
@@ -177,25 +177,6 @@ class Amplitude {
     }
   }
 
-  connect(unit) {
-    if (unit) {
-      if (unit.hasOwnProperty('input')) {
-        this.output.connect(unit.input);
-      } else {
-        this.output.connect(unit);
-      }
-    }
-    // else {
-    //   this.output.connect(this.panner.connect(p5sound.input));
-    // }
-  }
-
-  disconnect() {
-    if (this.output) {
-      this.output.disconnect();
-    }
-  }
-
   /**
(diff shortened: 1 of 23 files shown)


davepagurek (approved):
Looks good, thanks for implementing this everywhere and adding tests!

PR #762, opened 2023-03-30
update line endings for windows

blechdom:


montoyamoraga (commented):
we are trying to make grunt happen on a windows machine,
this PR was a mistake ~

